PR #59, opened 2015-01-12
json codec for publish payload

brharrington:
Supports legacy formats and field labels that are
still needed internally. It will likely be a long time
before we can fully deprecate and simplify those
paths.

Also has some to make it more efficient when working
with SmallHashMap values.


matschaffer:
Was bummed to notice that this doesn't include a WebApi. Is there any updated plan for some rudimentary publishing functionality?

brharrington:
I'm hoping to get it done by end of the week.

matschaffer:
:joy:

PR #62, opened 2015-01-14
support legacy offset

brharrington:
Trying to minimize internal vs external differences. This
adds support for the legacy offset operator until we can
fully deprecate and remove it.


matschaffer:
Curious, what's the rationale behind deprecating it? I found the new format a little cumbersome. Granted could just be that I'm used to the old way.

brharrington:
In general when crafting an expression there is a flow: query > aggregate > math > presentation. Originally there was just the offset that took the list and it is part of the presentation attributes, but ultimately needs to change the data fetched by the query. The second form was added so that math could be done on the results, e.g., subtract WoW from current line. We have considered the list version deprecated for a long time, but haven't really pushed the cleanup. 

The main reasons for deprecating it are to make it more consistent and reduce confusion from having the overload. I still get quite a few questions as most users don't understand the difference and the quirks of each. The most common use case with shifts is for the current time and a single shifted line like WoW. The overhead is about the same for that use-case:

* `name,sps,:eq,:sum,:dup,1w,:offset`
* `name,sps,:eq,:sum,(,0h,1w,),:offset`

For the list form common questions:

* How to show current time? The need to do an explicit shift of 0 isn't very intuitive. I have found when explaining it having a separate expression with a single shift is easier to understand.
* How to apply different presentation settings to each line? With the list form they are connected and so you have to essentially use it with a single item in the list to have different settings. So you wind up with something like: `name,sps,:eq,:sum,:dup,(,1w,),:offset`.

Do you have an example use-case that is easier with the list form?

matschaffer:
I had a bit of trouble when trying to apply a legend to the new form and ended up using this:

```
(,0h,1w,),(,name,sps,:eq,nf.cluster,nccp-silverlight,:eq,:and,:sum,:swap,:offset,$(name) $(atlas.offset),:legend,),:each
```

Maybe the `:rot,...,:legend` approach would have been easier though the separation of the query and legend info never quite sat right with me there.

matschaffer:
I hear you on the math justification though. I've seen that as a pain point in the past so probably a good move to get offset into the math stage.

brharrington:
I would typically do it more like:

```
name,sps,:eq,nf.cluster,nccp-silverlight,:eq,:and,:sum,
:dup,1w,:offset,
:list,(,$(name)+$(atlas.offset),:legend,),:each
```

In general I haven't seem that many use-cases where more than one shift is used on the same graph, so typically I only need to apply the offset to one line at the end. Since I frequently bind a cluster using `:cq` in the each expression it is fairly convenient.

matschaffer:
Oh that's much nicer. Would be good to have https://github.com/Netflix/atlas/wiki/Examples#time-shift show that as well I think. Seems like that's where I got the `list,list,:each` approach from.

brharrington:
Thanks, I haven't actually looked at some of the examples in a while. I agree it should be updated to show the more common use-case.

PR #75, opened 2015-01-17
initial import of publish webapi

brharrington:
Includes sample conf and a test script to show
and example of how to send data. To try it:

```
$ cd $REPO_DIR
$ sbt one-jar
$ java -jar atlas-webapi/target/atlas-webapi-1.4.2-SNAPSHOT-one-jar.jar conf/memory.conf
```

From another terminal:

```
$ scripts/publish-test.sh
```


matschaffer:
Like Christmas in January! Thanks so much for this!

PR #149, opened 2015-05-08
add 20s to list of auto-selected step sizes

brharrington:


brharrington:
Failures are due to some expected diffs where the auto-consolidated step hit 20m instead of 30m.

Before:

golden

After:

test

Diff:

diff

PR #208, opened 2015-07-31
add nf.job and nf.task keys

brharrington:
There is increasing interest in have a more common tagging
scheme for hadoop, spark, mantis, etc. All of these have
a notion of a job which gets broken down into a set of tasks.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/model/TagKey.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/model/TagKey.scala
@@ -75,6 +75,12 @@ object TagKey {
   /** Image id for the instance reporting the metric. */
   final val image = netflixPrefix + "ami"
 
+  /** Name for a job running in a framework like hadoop. */
+  final val job = netflixPrefix + "job"
+
+  /** Id for a task that runs as part of a job.  */
+  final val task = netflixPrefix + "task"
+
   /** ISO country code. */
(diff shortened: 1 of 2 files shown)


matschaffer:
First I'd realized that there were nf.* tags specified in the code. Is there info somewhere on how their use is intended/restricted?

brharrington:
We use them internally to have common concepts with agreed upon interpretation and avoid clashes with user defined tags. Specifically for data getting sent in we don't allow arbitrary tags in the `atlas.` and `nf.` namespaces. The `atlas.` is mostly for internal details for atlas the primary being how to treat the value for normalization and consolidation operations.

The `nf.` is a prefix we reserved early on for Netflix infrastructure use and they are typically not set by the user but automatically by the plugin when running in our environment. The publish backend will reject data that includes unknown `nf.` prefixed keys and we can potentially do additional validation such as verify that nf.app values are defined in Asgard. We restrict to avoid arbitrary clutter for that category and also so we can safely add/change the set in the future without clashing with user defined tags.

That said, for the `nf.` prefix they will likely get refactored out into configuration rather than being in the code, but it isn't a high priority. This is the first update in several years so it just hasn't gotten much attention. You can already choose to remove that validation rule or create a custom one to change this behavior.

matschaffer:
Ah, that makes sense. Would be nice to have something configurable eventually but I could see why it's not a big priority. Thanks for the explanation!

PR #222, opened 2015-09-09
experimental sliding des operation

brharrington:
This is still experimental and being added in so we
can test the impact.

One of the common complaints with DES is that to get
the same value for a given time you must start feeding
in data at exactly the same time. So for normal graphs
where it is computed using the window of the chart it
will have slightly different predictions for a given
time. As it is often used for alerting this makes it
cumbersome to try and determine:

1. Why an alarm fired
2. When alarms would have fired for tuning

For these use-cases a model that is deterministic within
a small window would be useful. Sliding DES uses two DES
functions and alternates between them. One will get trained
while the other is getting used, and then the one that was
getting used will get reset and the roles swapped. If the
windows are aligned based on the training size this
provides a simple way to make a DES based prediction
that is deterministic for arbitrary windows.

```
 F1 | A |-- T1 --|-- P1 --|-- T1 --|-- P1 --|-- T1 --|
 F2 | A |        |-- T2 --|-- P2 --|-- T2 --|-- P2 --|

Result:

 R  |-- NaN -----|-- P1 --|-- P2 --|-- P1 --|-- P2 --|

```

Both functions will ignore any data until it reaches a
boundary, even multiple, of the training window. That is
shown as `A` in the diagram above. The first function will
then start training, and after the training window the first
predicted values will get generated.

One side-effect of this approach is that there can be abrupt
transitions when switching from one function to another.
Overall it is not expected to hurt the general usage, but
is visually less appealing. Also, for actual predictions it
seems to help in some cases because the transition can
cause the impact of sharp changes to correct more quickly.

For now, we are adding in a separate `:sliding-des` operator
that can be used for testing with real data. Depending on the
outcome we may change the existing `:des` function to be
`:sliding-des`.


brharrington:
Examples from wiki of sharp and smooth drops. The area shows the current `:des` compared to the `:sliding-des` in red. Smooth example in particular is quite jagged during transitions.

sharp-drop

smooth-drop

PR #258, opened 2015-12-09
index for efficiently matching many queries

brharrington:
The current indexes allow us to quickly check a query
against many tag sets. For some use-cases, in particular,
stream processing, we need to check a tag set against
many queries.

The specific use-case is inline alert processing. Currently
in the main environment there are around 17k alert expressions
used for alerts. Decomposing those into individual queries
we get around 33k queries. With a quick test of comparing
a real payload of 5000 datapoints to find matches within the
alert query set takes around 6.2s using a naive loop. This
query index takes around 6ms.

We'll likely need to do some further optimization to reduce
the number of allocations when doing the match, but current
testing shows it to be good enough to include in a snapshot
for more experimentation.


brharrington:
@dmuino can you review?

PR #291, opened 2016-01-18
add :percentiles and :median

brharrington:
Adds a `:percentiles` helper to compute estimated percentiles
for spectator PercentileTimer and PercentileDistributionSummary.
These use a precomputed set of buckets that partition the range
of non-negative long values to allow for arbitrary slicing and
dicing based on the dimensions and still get a reasonable
approximation of percentiles.

This approach is less accurate than alternatives like t-Digest,
but more efficent and practical for our use-cases.

It is used just after the data function which should be a sum
or group by:

```
name,requestLatency,:eq,(,50,90,),:percentiles
name,requestLatency,:eq,(,nf.cluster,),:by,(,50,90,),:percentiles
```

The output will be one line per input per requested percentile.
There is also a helper, `:median` that is just short hand for
`(,50,),:percentiles`.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/model/MathVocabulary.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/model/MathVocabulary.scala
@@ -49,6 +49,8 @@ object MathVocabulary extends Vocabulary {
 
     Sum, Count, Min, Max,
 
+    Percentiles,
+
     Macro("avg", List(":dup", ":sum", ":swap", ":count", ":div"), List("name,sps,:eq,(,nf.cluster,),:by")),
 
     Macro("pct", List(":dup", ":sum", ":div", "100", ":mul"), List("name,sps,:eq,(,nf.cluster,),:by")),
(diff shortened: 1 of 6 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/model/MathVocabulary.scala
```diff
@@ -641,4 +645,54 @@ object MathVocabulary extends Vocabulary {
...
+    }
+
+    private def toSum(af: AggregateFunction): DataExpr.Sum = {
+      DataExpr.Sum(af.query, offset = af.offset)
+    }
+
+    override def summary: String =
+      """
+        |Estimate percentiles for a timer or distribution summary. The data must have been
+        |published with appropriately to allow the approximation. If using
```
dmuino: Replace "published with appropriately" with "published with appropriate values for the required percentile tag"?

Not sure if going into detail about the T and D prefix plus 4 hex digit convention is appropriate here. Maybe also explaining T means dividing by 1e9.

brharrington: > Not sure if going into detail about the T and D prefix plus 4 hex digit convention is appropriate here. 

In most cases users shouldn't need to know about that, they just need to know to use the appropriate type from spectator or other libraries. In a separate PR I'll update the atlas/spectator wikis with more details on how we are encoding it, but users shouldn't need to know.

PR #343, opened 2016-04-15
docs: create search index for wiki

brharrington:
Generate search index json file for the wiki
that can be consumed easily with lunr. Format
is similar to index created by mkdocs, e.g.:

https://netflix.github.io/spectator/en/latest/mkdocs/search_index.json


brharrington:
https://github.com/Netflix/atlas/wiki/search_index.json

PR #380, opened 2016-06-18
hex decoding for legend values (bug 1271)

brharrington:
Provides a style operator, `:decode`, that can be
used to perform hex decoding of the legend text.

Background: internally we restrict input to simple
alpha-numeric characters, `.`, `-`, and `_`. if
characters are used the data will get dropped as
invalid. If using the servo-atlas module, then it
will convert invalid characters to `_`. Most
commonly that was done for spaces.

The restriction is to ensure that it is easy and
safe to craft the query expressions and include
them as part of the URI in a variety of contexts.
Precise linkability is extremely important and
should not be compromised.

Some users do their own url encoding and the `%`
then gets converted to an `_`. This allows them
to undo the restriction and get the original
characters back. With this change that decoding
can now be specified as part of the expression.

There are quite a few misgivings about this
feature. In my view this feature encourages bad
behavior such as trying to encode structural
information into tag values or trying to abuse
it as a more general purpose data store. The
compromise for now is supporting the decode
with a clear warning in the docs and we'll see
how it goes.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/model/StyleExpr.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/model/StyleExpr.scala
@@ -30,7 +30,12 @@ case class StyleExpr(expr: TimeSeriesExpr, settings: Map[String, String]) extend
 
   def legend(t: TimeSeries): String = {
     val fmt = settings.getOrElse("legend", t.label)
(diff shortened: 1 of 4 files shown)


brharrington:
Note, wiki update was already run: https://github.com/Netflix/atlas/wiki/style-decode

PR #382, opened 2016-07-02
Sdes streaming

skandragon:
Fix :sdes to return identical values if passed as a single batch or incrementally.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/model/StatefulExpr.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/model/StatefulExpr.scala
@@ -143,7 +143,7 @@ object StatefulExpr {
       trainingSize: Int,
       alpha: Double,
       beta: Double) extends StatefulExpr {
-    import com.netflix.atlas.core.model.StatefulExpr.Des._
+    import com.netflix.atlas.core.model.StatefulExpr.SlidingDes._
 
     def dataExprs: List[DataExpr] = expr.dataExprs
     override def toString: String = s"$expr,$trainingSize,$alpha,$beta,:sdes"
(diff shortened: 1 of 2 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/model/StatefulExpr.scala
```diff
@@ -152,26 +152,30 @@ object StatefulExpr {
...
     def groupByKey(tags: Map[String, String]): Option[String] = expr.groupByKey(tags)
 
-    private def eval(ts: ArrayTimeSeq, s: State, skip: Int): State = {
-      val desF = new OnlineSlidingDes(trainingSize, alpha, beta)
-      desF.reset()
-      desF.currentSample = s.currentSample
+    private def eval(ts: ArrayTimeSeq, s: State, skip: Long): State = {
+      var desF : OnlineSlidingDes = s.desF
+      var skipUpTo = s.skipUpTo
+      if (desF == null) {
```
brharrington: It seems like this check could be handled by the else block of `state.getOrElse` on line 193. Simply initialize the state with the correct step and des function.

atlas-core/src/main/scala/com/netflix/atlas/core/model/StatefulExpr.scala
```diff
@@ -182,21 +186,20 @@ object StatefulExpr {
...
+        state(t.id) = eval(bounded, s, getAlignedStartTime(context))
         TimeSeries(t.tags, s"sdes(${t.label})", bounded)
       }
       ResultSet(this, newData, rs.state + (this -> state))
     }
   }
 
   object SlidingDes {
-    case class State(pos: Int, currentSample: Int, sp: Double, bp: Double)
+    case class State(skipUpTo: Long, desF: OnlineSlidingDes)
```
brharrington: Thinking about this some more, right now state is immutable and with this change we lose the immutability. 

```scala
val r1 = expr.eval(context, data)
val r2 = expr.eval(context, data)
assert(r1 === r2) // Fails because state within the context was mutated
```

atlas-core/src/test/scala/com/netflix/atlas/core/model/DesSuite.scala
```diff
@@ -38,10 +39,27 @@ class DesSuite extends FunSuite {
     List(Datapoint(dataTags, 12L * step, 1.2)),
     List(Datapoint(dataTags, 13L * step, 1.2))
   )
-
-  val inputTS = TimeSeries(dataTags, new ArrayTimeSeq(DsType.Gauge, 0L, step,
+  val allignedInputTS = TimeSeries(dataTags, new ArrayTimeSeq(DsType.Gauge, 0L, step,
```
brharrington: s/alligned/aligned/

atlas-core/src/test/scala/com/netflix/atlas/core/model/DesSuite.scala
```diff
@@ -38,10 +39,27 @@ class DesSuite extends FunSuite {
...
     Array[Double](1.0, 1.5, 1.6, 1.7, 1.4, 1.3, 1.2, 1.0, 0.0, 0.0, 1.0, 1.1, 1.2, 1.2)))
 
+  val unallignedStream = List(
```
brharrington: s/unalligned/unaligned/

atlas-core/src/main/scala/com/netflix/atlas/core/model/StatefulExpr.scala
skandragon: I think we decided to leave it this way for now, but to refactor underlying objects to be able to save state and construct from saved state.  This will be done later.

PR #389, opened 2016-07-07
Export state

skandragon:
For future work, we will want to render the internal state to JSON or similar.  This change set adds a side object for OnlineDes and OnlineSlidingDes which is a simple case class, easily rendered to JSON.  This state class is used in the :des and :sdes operators to store and restore internal state, which allows this to be rendered to JSON as well.

Additionally, the state passed into StatefulExpression eval() methods is not modified in any expression.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineDes.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineDes.scala
@@ -28,16 +28,9 @@ package com.netflix.atlas.core.algorithm
 class OnlineDes(training: Int, alpha: Double, beta: Double) {
   import OnlineDes._
 
-  var currentSample = 0
-  var sp = Double.NaN
-  var bp = Double.NaN
-
-  def this(state: OnlineDes.State) = {
-    this(state.training, state.alpha, state.beta)
-    currentSample = state.currentSample
-    sp = state.sp
-    bp = state.bp
-  }
+  private var currentSample = 0
+  private var sp = Double.NaN
+  private var bp = Double.NaN
 
   def next(v: Double): Double = {
     val retval = if (currentSample >= training) sp else Double.NaN
@@ -61,12 +54,19 @@ class OnlineDes(training: Int, alpha: Double, beta: Double) {
     bp = Double.NaN
   }
 
-  def state : State = State(training, alpha, beta, currentSample, sp, bp)
+  def state: State = State(training, alpha, beta, currentSample, sp, bp)
 }
 
 object OnlineDes {
   case class State(training: Int, alpha: Double, beta: Double, currentSample: Int, sp: Double, bp: Double)
 
-  def apply(state: State) = new OnlineDes(state)
-  def apply(training: Int, alpha: Double, beta: Double) = new OnlineDes(training, alpha, beta)
+  def apply(state: State) : OnlineDes = {
+    var des = new OnlineDes(state.training, state.alpha, state.beta)
+    des.currentSample = state.currentSample
+    des.sp = state.sp
+    des.bp = state.bp
+    des
+  }
+
+  def apply(training: Int, alpha: Double, beta: Double): OnlineDes = new OnlineDes(training, alpha, beta)
 }
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineDes.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineDes.scala
@@ -26,10 +26,18 @@ package com.netflix.atlas.core.algorithm
  *     Trend smoothing factor.
  */
 class OnlineDes(training: Int, alpha: Double, beta: Double) {
+  import com.netflix.atlas.core.algorithm.OnlineDes._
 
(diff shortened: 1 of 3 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineDes.scala
```diff
@@ -26,11 +26,19 @@ package com.netflix.atlas.core.algorithm
  *     Trend smoothing factor.
  */
 class OnlineDes(training: Int, alpha: Double, beta: Double) {
+  import com.netflix.atlas.core.algorithm.OnlineDes._
```
dmuino: You don't need to qualify OnlineDes here. A simple:

```import OnlineDes._```

suffices.

File: atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineSlidingDes.scala [hunk 2]
```diff
@@ -26,13 +26,22 @@ package com.netflix.atlas.core.algorithm
  * @param beta
  *     Trend smoothing factor.
  */
-class OnlineSlidingDes(training: Int, alpha: Double, beta: Double) {
+class OnlineSlidingDes(training: Int, alpha: Double, beta: Double, des1: OnlineDes, des2: OnlineDes) {
+  import com.netflix.atlas.core.algorithm.OnlineSlidingDes._
```
dmuino: ```import OnlineSlideDes._```

skandragon: I was attempting to match the style of explicit imports found elsewhere, such as the StatefulExpr.scala file:

```  case class RollingCount(expr: TimeSeriesExpr, n: Int) extends StatefulExpr {
    import com.netflix.atlas.core.model.StatefulExpr.RollingCount._
```

I'm happy to shorten it, but perhaps we should consider doing that globally?

brharrington: > I'm happy to shorten it, but perhaps we should consider doing that globally?

I'm not as picky about that specifically. It is fine with me if you want to send a PR that makes them all consistently shorter.

atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineDes.scala
```diff
@@ -52,4 +60,13 @@ class OnlineDes(training: Int, alpha: Double, beta: Double) {
     sp = Double.NaN
     bp = Double.NaN
   }
+
+  def state : State = State(training, alpha, beta, currentSample, sp, bp)
```
brharrington: s/state : State/state: State/

atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineDes.scala
```diff
@@ -26,11 +26,19 @@ package com.netflix.atlas.core.algorithm
...
   var currentSample = 0
   var sp = Double.NaN
   var bp = Double.NaN
 
+  def this(state: OnlineDes.State) = {
+    this(state.training, state.alpha, state.beta)
+    currentSample = state.currentSample
+    sp = state.sp
+    bp = state.bp
+  }
```
brharrington: I think you could get rid of this constructor and just use `apply` method in the object.

atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineDes.scala
```diff
@@ -52,4 +60,13 @@ class OnlineDes(training: Int, alpha: Double, beta: Double) {
...
     bp = Double.NaN
   }
+
+  def state : State = State(training, alpha, beta, currentSample, sp, bp)
+}
+
+object OnlineDes {
+  case class State(training: Int, alpha: Double, beta: Double, currentSample: Int, sp: Double, bp: Double)
+
+  def apply(state: State) = new OnlineDes(state)
```
brharrington: You should be able to get rid of the second constructor and do most of the work in this method, e.g.:

```scala
def apply(state: State): OnlineDes = {
  val des = new OnlineDes(state.training, state.alpha, state.beta)
  des.currentSample = state.currentSample
  des.sp = state.sp
  des.bp = state.bp
  des
}
```

Side note, I generally prefer to be explicit about the return type on public methods. You will get it via inference, but change to that can create binary incompatible changes.

atlas-core/src/main/scala/com/netflix/atlas/core/model/StatefulExpr.scala
```diff
@@ -114,17 +119,19 @@ object StatefulExpr {
         data(pos) = desF.next(yn)
         pos += 1
       }
-      State(0, desF.currentSample, desF.sp, desF.bp)
+      State(0, desF.state)
```
brharrington: Is the `pos` field still needed in the state? It doesn't look to me like it is actually getting used.

skandragon:
I believe all the changes suggested are complete.  I also made some of the members private, as they can be now.

PR #407, opened 2016-08-06
reduce allocations for computing ids

brharrington:
Computing the id for a tag map is in the hot path when publishing metrics. It is currently the biggest chunk of allocations on that path and it looks like there are a few simple changes that would reduce it significantly.


brharrington:
Allocations before any changes:

before_buffer_change

brharrington:
After 5811f66:

after_reuse_buffer

PR #411, opened 2016-08-09
switch to MurmurHash3.arrayHash

brharrington:
Switches the hashCode implementation for SmallHashMap
to call `MurmurHash3.arrayHash` instead of relying on
the implementation from the super class which calls
`MurmurHash3.mapHash`. The array variant is much faster
(3-4x) and greatly reduces allocations.


brharrington:
Allocations for mapHash:

screen shot 2016-08-09 at 4 06 07 pm

Allocations for arrayHash:

screen shot 2016-08-09 at 4 06 24 pm

PR #417, opened 2016-08-16
Set actor name rather than using a generated one.

skandragon:
Without this change, automatically generated paths are made (such as ```akka://atlas/user/$a```) rather than what we specified in our config.  This change calls the explicit-name version of ```actorOf()``` to fix this.

As a side note, in travis, we should probably fire up the server once and verify we can post something to it.  All the tests pass without this change, but the stand-alone server could not find actors as the names were wrong.


brharrington:
Thanks, the change LGTM. It appears to have been a recent regression from #415. I'm a bit surprised none of the existing tests caught this, I'll double check why that is and make sure we get test coverage there.

PR #418, opened 2016-08-18
manager for polling data

brharrington:
Adds in the base manager that supervises a set of poller
actors. This is used for collecting metrics data from
various other services such as edda, healthcheck service,
cloudwatch, etc.


pparth:
@brharrington  this looks interesting but i cannot grasp the context. Can you provide more information or documentation?

PR #421, opened 2016-08-23
add poller to collect CloudWatch metrics data

brharrington:
Poller for collecting CloudWatch data and publishing
it into Atlas. This is a restructured from the legacy
db implementation used internally to wrap CloudWatch.
Overall the polling approach is preferred because:

* It makes CloudWatch data follow the same path as
  all other data. This means it is available for teeing
  to stream processing systems, long term storage in
  hive, etc.
* Querying CloudWatch on-demand does not scale well
  and it fails frequently due to throttles. In this
  case there is a consistent load all the time, but
  we can tune that for the needed set and avoid the
  really high bursts needed for on-demand queries.
* If we are lucky Amazon will support a way to get a
  stream of all CloudWatch metric data and we can
  switch to consuming that in the future. In the
  mean time we can make it look like a stream for
  most of our use-cases.

### DIFF
--- a/atlas-module-cloudwatch/src/main/resources/META-INF/services/com.google.inject.Module
+++ b/atlas-module-cloudwatch/src/main/resources/META-INF/services/com.google.inject.Module
@@ -1,0 +1,1 @@
+com.netflix.atlas.guice.CloudWatchModule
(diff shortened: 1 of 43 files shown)


pparth:
Is there any documentation regarding this Cloudwatch support feature?

brharrington:
Not yet, I've got a couple documentation tasks on the todo list.

PR #431, opened 2016-09-01
reduce size of publish failure response

brharrington:
Fixes a bug where the failure message was getting encoded as
a json string inside the message of the json response. Also
changes the behavior on failures to have the error count
explicitly in the response and limit the number of specific
failure messages to the first 5. This significantly reduces
allocations and response sizes for payloads with many failures.


brharrington:
Backport of #409 to fix https://github.com/spring-cloud/spring-cloud-netflix/issues/798.

PR #453, opened 2016-11-08
some patterns incorrectly interpretted as indexOf

brharrington:
For simple patterns the `StringMatcher` will convert
them to basic string operations that perform better
than using regular expressions.

If a simple pattern had an explicit end anchor, then it
was incorrectly choosing the `IndexOf` matcher rather
than falling back to using `Regex`. This change fixes
the `IndexOf` pattern and also adds an `Equals` matcher
to cover the simple case using `String.equals`.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/util/StringMatcher.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/util/StringMatcher.scala
@@ -28,23 +28,28 @@ object StringMatcher {
   /** Matches everything. */
   private val AllPattern = """^\^*(\.\*)*\$*$""".r
 
+  /** A pattern that can be checked with the String.equals method. */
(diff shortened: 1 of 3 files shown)


brharrington:
/cc @dmuino @svachalek 

**Before**

```
$ curl -s 'http://localhost:7101/api/v1/tags/name?format=txt&q=name,s$,:re'
DiscoveryStatus_DOWN
DiscoveryStatus_UP
DiscoveryStatus_nccp_UP
playback.startLatency
poller.asg.instance
requestsPerSecond
sps
ssCpuUser
```

**After**

```
$ curl -s 'http://localhost:7101/api/v1/tags/name?format=txt&q=name,s$,:re'
$ curl -s 'http://localhost:7101/api/v1/tags/name?format=txt&q=name,sps$,:re'
sps
```

PR #459, opened 2016-11-21
Lwcapi merge

skandragon:
Merge in lwcapi.  This adds the requirement for travis to run a redid-server, although that is intended to go away in the future.


atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
```diff
@@ -0,0 +1,118 @@
...
+class SSEActor(client: ActorRef,
+               sseId: String,
+               name: String,
+               sm: SubscriptionManager,
+               registry: Registry)
+  extends Actor with ActorLogging {
+  import SSEActor._
+  import StreamApi._
+
+  private val connectsId = registry.createId("atlas.lwcapi.sse.connectCount")
```
brharrington: The naming here is a bit inconsistent with that used for redis connects (`atlas.lwcapi.redis.connects` vs `atlas.lwcapi.sse.connectCount`).

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
```diff
@@ -0,0 +1,118 @@
...
+  private val droppedId = registry.createId("atlas.lwcapi.sse.droppedCount")
+
+  private val sseCountId = registry.createId("atlas.lwcapi.streams").withTag("streamType", "sse")
+  private val sseCount = registry.gauge(sseCountId, new AtomicInteger(1))
+
+  private var outstandingCount = 0
+  private val maxOutstanding = 100
+  private var droppedCount = 0
+
+  private val instanceId = sys.env.getOrElse("EC2_INSTANCE_ID", "unknown")
```
brharrington: See earlier comment on NetflixEnvironment.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
```diff
@@ -0,0 +1,118 @@
...
+  override def postStop() = {
+    unregister()
+    ticker.cancel()
+    super.postStop()
+  }
+}
+
+object SSEActor {
+  case class Ack()
+  case class Tick()
```
brharrington: Use `case object` instead. You don't need to create an instance for each use.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
```diff
@@ -0,0 +1,118 @@
...
+  var ticker: Cancellable = context.system.scheduler.scheduleOnce(tickTime) {
+    self ! Tick()
+  }
+
+  def receive = {
+    case Ack() =>
+      outstandingCount -= 1
+    case Tick() =>
+      if (outstandingCount == 0) send(tickMessage)
+      ticker = context.system.scheduler.scheduleOnce(tickTime) {
```
brharrington: Why not just schedule repeatedly rather than once each time?

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
```diff
@@ -0,0 +1,118 @@
...
+      log.info(s"SSE Stream closed: $closed")
+      context.stop(self)
+  }
+
+  private def send(msg: SSEMessage): Unit = {
+    if (outstandingCount < maxOutstanding) {
+      val json = msg.toSSE + "\r\n\r\n"
+      client ! MessageChunk(json).withAck(Ack())
+      outstandingCount += 1
+      registry.counter(messagesId.withTag("action", msg.getWhat).withTag("streamId", sseId)).increment()
```
brharrington: Wouldn't there be a lot of stream ids that would change frequently? It doesn't seem appropriate for a tag value.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
```diff
@@ -0,0 +1,118 @@
...
+    } else {
+      droppedCount += 1
+      registry.counter(droppedId.withTag("action", msg.getWhat).withTag("streamId", sseId)).increment()
```
brharrington: It would be a little more efficient to do: `withTags("action", msg.getWhat, "streamId", sseId)`.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/StreamApi.scala
```diff
@@ -0,0 +1,138 @@
...
+  def routes: RequestContext => Unit = {
+    path("lwc" / "api" / "v1" / "stream" / Segment) { (streamId) =>
+      parameters('name.?, 'expression.?, 'frequency.?) { (name, expr, frequency) =>
+        get { (ctx) => handleReq(ctx, streamId, name, expr, frequency) } ~
+        post { (ctx) => handleReq(ctx, streamId, name, expr, frequency) }
+      }
+    }
+  }
+
+  private def handleReq(ctx: RequestContext, streamId: String, name: Option[String], expr: Option[String], freqString: Option[String]): Unit = {
```
brharrington: Try to wrap at 100 chars.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/StreamApi.scala
```diff
@@ -0,0 +1,138 @@
...
+  import StreamApi._
+
+  private val dbActor = actorRefFactory.actorSelection("/user/lwc.expressiondb")
+  private val subscribeRef = actorRefFactory.actorSelection("/user/lwc.subscribe")
+
+  def routes: RequestContext => Unit = {
+    path("lwc" / "api" / "v1" / "stream" / Segment) { (streamId) =>
+      parameters('name.?, 'expression.?, 'frequency.?) { (name, expr, frequency) =>
+        get { (ctx) => handleReq(ctx, streamId, name, expr, frequency) } ~
+        post { (ctx) => handleReq(ctx, streamId, name, expr, frequency) }
```
brharrington: A post that takes query params on the URL seems a bit odd.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/TTLManager.scala
```diff
@@ -0,0 +1,60 @@
...
+      val key = next.get._2
+      val item = lastTouched(key)
+      if (item == 0) {
+        // we don't care about this one, drop it
+        prio.dequeue
+        needsTouch(since)
+      } else if (-item != next.get._1) {
+        // requeue with the latest timestamp
+        prio.dequeue
+        prio += ((-item, key))
```
brharrington:
```
(-item -> key)
```

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/TTLManager.scala
```diff
@@ -0,0 +1,60 @@
...
+  private val touched = mutable.Map[K, Long]()
+
+  def lastTouched(key: K): Long = {
+    touched.getOrElse(key, 0)
+  }
+
+  def touch(key: K, now: Long): Unit = {
+    val old = touched.put(key, now)
+    if (old.isEmpty)
+      prio += ((-now, key))
```
brharrington: Add comment about why this is negated.

build.sbt
```diff
@@ -66,6 +68,17 @@ lazy val `atlas-json` = project
     Dependencies.jodaConvert
   ))
 
+lazy val `atlas-lwcapi` = project
+  .configure(BuildSettings.profile)
+  .dependsOn(`atlas-akka`, `atlas-core`, `atlas-json`, `atlas-test` % "test")
+  .settings(libraryDependencies ++= Seq(
+    Dependencies.frigga,
+    Dependencies.spectatorSandbox,
```
skandragon: It's probably not used at this point, but was somewhere in the experimentation phase.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/EvaluateActor.scala
```diff
@@ -0,0 +1,58 @@
...
+import com.netflix.spectator.api.Registry
+import spray.http.{HttpResponse, StatusCodes}
+
+class EvaluateActor @Inject() (registry: Registry, sm: SubscriptionManager) extends Actor with ActorLogging {
+  import com.netflix.atlas.lwcapi.EvaluateApi._
+
+  private val evalsId = registry.createId("atlas.lwcapi.evaluate.count")
+  private val itemsId = registry.createId("atlas.lwcapi.evaluate.itemCount")
+  private val actorsId = registry.createId("atlas.lwcapi.evaluate.actorCount")
+  private val uninterestingId = registry.createId("atlas.lwcapi.evaluate.uninterestingCount")
```
skandragon: Will fix.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionApi.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import akka.actor.ActorRefFactory
+import com.netflix.atlas.akka.WebApi
+import com.netflix.atlas.json.{Json, JsonSupport}
+import com.netflix.spectator.api.Registry
+import com.typesafe.scalalogging.StrictLogging
+import spray.http.{HttpResponse, StatusCodes}
+import spray.routing.RequestContext
+
+case class ExpressionApi @Inject()(alertmap: ExpressionDatabase,
```
skandragon: Renaming issue from a long time ago.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionDatabaseActor.scala
```diff
@@ -0,0 +1,357 @@
...
+  private val uuid = GlobalUUID.get
+
+  private val ttl = ApiSettings.redisTTL
+  private val host = ApiSettings.redisHost
+  private val port = ApiSettings.redisPort
+
+  private val redisCmdHeartbeat = "hb"
+  private val redisCmdExpression = "expr"
+
+  private val heartbeatInterval = 10000 // milliseconds
```
skandragon: I thought about that, but in general I don't think it's useful to have it configurable.  These are internal only -- these are not exposed to the caller, and the heartbeat is used just to ensure we detect a dead redis connection more than anything else.

conf/lwcapi-2.conf
```diff
@@ -0,0 +1,15 @@
...
+      ttl = 20000
+    }
+  }
+
+  akka {
+    port = 7102
+    actors = ${?atlas.akka.actors}
+    api-endpoints = ${?atlas.akka.api-endpoints}
+  }
+}
```
skandragon: https://github.com/typesafehub/config/blob/master/HOCON.md#include-semantics-locating-resources seems to imply it'll work, so I'll change this.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/DatabaseService.scala
```diff
@@ -0,0 +1,38 @@
...
+import javax.inject.Singleton
+
+import com.netflix.iep.service.{AbstractService, State}
+import com.typesafe.scalalogging.StrictLogging
+
+@Singleton
+class DatabaseService extends AbstractService with StrictLogging {
+  @volatile private var started = false
+
+  def setState(value: Boolean): Unit = started = value
```
skandragon: I thought "state" was the state of the registration, while "started" really means "initial transfer from other instances via redis is complete"

I can rename things to to make it clearer, and add commends, unless I really can / should touch "state" here.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionDatabaseActor.scala
```diff
@@ -0,0 +1,357 @@
...
+  }
+
+  def processRedisExpression(cmd: String, originator: String, json: String, now: Long) = {
+    val req = RedisExpressionRequest.fromJson(json)
+    logger.debug(s"pubsub add for expressionId ${req.id}")
+    logRedisCommand(cmd, originator, "redisReceive", req)
+    if (!alertmap.hasExpr(req.id)) {
+      val split = splitter.split(req.expression, req.frequency)
+      if (split.queries.size != 1) {
+        logger.error(s"Redis found more than one data expression: ${split.queries.size} ${req.expression}")
```
skandragon: While this if statement should never trigger, if someone changes the code later and this doesn't hold true, it's better to catch it here I think, with a useful logging message.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionDatabaseActor.scala
```diff
@@ -0,0 +1,357 @@
...
+      split.queries.zip(split.expressions).foreach { case (query, expr) =>
+        alertmap.addExpr(expr, query)
+        redisPublish(RedisExpressionRequest(expr.id, expr.expression, expr.frequency))
+      }
+    case Subscribe(streamId, expressionId) =>
+      logger.debug(s"Adding sub streamId $streamId expressionID $expressionId")
+      increment_counter(updatesId, "local", actionSubscribe)
+      sm.subscribe(streamId, expressionId)
+    case Unsubscribe(streamId, expressionId) =>
+      logger.debug(s"Adding unsub streamId $streamId expressionID $expressionId")
```
skandragon: Heh.  Yes.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionDatabaseActor.scala
```diff
@@ -0,0 +1,357 @@
...
+    while (!done) {
+      val top = ttlManager.needsTouch(targetTime)
+      if (top.isDefined) {
+        logger.debug(s"TTL: Checking ${top.get}")
+        touchExpression(top.get, now)
+      } else {
+        done = true
+      }
+    }
+  }
```
skandragon: Used.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionDatabaseActor.scala
```diff
@@ -0,0 +1,357 @@
...
+
+  def touchExpression(id: String, now: Long) = {
+    val state = ttlState(id)
+    state match {
+      case TTLState.NotPresent => // do nothing
+        logger.warn(s"TTL: Expression state for $id is strangely NotPresent")
+      case TTLState.Active =>
+        val subscriberPresent = sm.actorsForExpression(id).nonEmpty
+        val split = alertmap.expr(id)
+        if (!subscriberPresent || split.isEmpty) {
```
skandragon: Comment added.  The situation is we may either have no subscribers (locally) or we may have deleted the Id from our database.  Either way, we want to move the TTL tracking state to PendingDelete, which means we will stop advertising it.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionDatabaseActor.scala
```diff
@@ -0,0 +1,357 @@
...
+    var tries = 1
+    while (true) {
+      try {
+        logger.info(s"Connecting to redis($source), tries $tries")
+
+        registry.counter(connectsId.withTag("source", source)).increment()
+        if (tries != 1) {
+          registry.counter(connectRetriesId.withTag("source", source)).increment()
+        }
+        Thread.sleep(1000)
```
skandragon: My hope is to replace this whole system.  For now, while sleep is messy -- this only happens when redis is down, which also means we are not working anyway.

I could convert this over to something better, but I'd rather convert it over to something that isn't Redis.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionDatabaseActor.scala
```diff
@@ -0,0 +1,357 @@
...
+    }
+    None
+  }
+
+  override def postStop() = {
+    ticker.cancel()
+    super.postStop()
+  }
+
+  case class TTLItem(flavor: String, id: String) extends Ordered[TTLItem] {
```
skandragon: It's not anymore.  Removed.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/EvaluateActor.scala
```diff
@@ -0,0 +1,58 @@
...
+    case EvaluateRequest(timestamp, items) =>
+      evaluate(timestamp, items)
+      sender() ! HttpResponse(StatusCodes.OK)
+    case _ =>
+      DiagnosticMessage.sendError(sender(), StatusCodes.BadRequest, "unknown payload")
+  }
+
+  private def evaluate(timestamp: Long, items: List[Item]): Unit = {
+    registry.counter(evalsId).increment()
+    registry.counter(itemsId).increment(items.size)
```
skandragon: Since I iterate over the list, I could increment a counter and update the metric at the end.  Do you think this is necessary?

brharrington: I'm not too worried about it in this particular case, just mentioning in case there is an easy way to avoid it. I have seen things like size calls become pretty expensive in the past and it is often assumed it will be a constant time operation. When we start running it under more load we can deal with it if it shows up on a profile.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionSplitterImpl.scala
```diff
@@ -0,0 +1,108 @@
...
+import com.netflix.atlas.core.model.Query.KeyQuery
+import com.netflix.atlas.core.model.{ModelExtractors, Query, StyleVocabulary}
+import com.netflix.atlas.core.stacklang.Interpreter
+
+class ExpressionSplitterImpl extends ExpressionSplitter {
+  import ExpressionSplitter._
+  private val keepKeys = Set("nf.app", "nf.stack", "nf.cluster")
+
+  private val interpreter = Interpreter(StyleVocabulary.allWords)
+  private val interner = scala.collection.mutable.AnyRefMap[Query, Query]()
```
skandragon: The interner is only modified from a synchronized split() method, so there should be no threading issues.  There may be a hot-spot however, but I'm not sure how to avoid this.

It is possible this list could grow, but the names of metrics and the such seem like it's practically bounded, even if it's not actually bounded.  Do you think we need to worry about this, and periodically swap out the interner to allow it to clean up?

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionSplitterImpl.scala
```diff
@@ -0,0 +1,108 @@
...
+      case ModelExtractors.PresentationType(t) =>
+        t.expr.dataExprs.map(e => intern(compress(e.query)))
+      case _ => throw new IllegalArgumentException("Expression is not a valid expression")
+    }
+    val expressions = distinctStack.flatMap {
+      case ModelExtractors.PresentationType(t) =>
+        t.expr.dataExprs.map(e => ExpressionWithFrequency(e.toString, frequency))
+      case _ => throw new IllegalArgumentException("Expression is not a valid expression")
+    }
+    SplitResult(queries, expressions)
```
skandragon: After a little futzing, I found a way, which passes tests. :)

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
skandragon: Ooops, that was to diagnose a problem.  Fixed (removed)

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
skandragon: fixed.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
skandragon: Fixed.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
skandragon: I added Count to the redis stuff, since I like more descriptive counters.

PR #475, opened 2017-01-04
track bytes sent, not messages sent, for limiting resource usage

skandragon:
This patch changes limiting outstanding messages to limiting outstanding bytes for slower readers.  The limit is set to 10 megabytes per stream.

Also, individual metrics are combined into a single HTTP write request every 100 ms (hard coded) or 100 k of buffered data (hard coded)

### DIFF
--- a/atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
+++ b/atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
@@ -37,47 +37,56 @@ class SSEActor(client: ActorRef,
 
   private val connectsId = registry.createId("atlas.lwcapi.sse.connectCount")
   private val messagesId = registry.createId("atlas.lwcapi.sse.messageCount")
-  private val droppedId = registry.createId("atlas.lwcapi.sse.droppedCount")
+  private val messageBytesId = registry.createId("atlas.lwcapi.sse.messageBytesCount")
+  private val sendsId = registry.createId("atlas.lwcapi.sse.httpSendsCount")
+  private val sendBytesId = registry.createId("atlas.lwcapi.sse.httpSendBytesCount")
+  private val droppedId = registry.createId("atlas.lwcapi.sse.droppedMessageCount")
 
   private val sseCountId = registry.createId("atlas.lwcapi.streams").withTag("streamType", "sse")
   private val sseCount = registry.gauge(sseCountId, new AtomicInteger(1))
 
-  private var outstandingCount = 0
-  private val maxOutstanding = 100
-  private var droppedCount = 0
+  private var outstandingByteCount: Long = 0
+  private val maxOutstandingBytes: Long = 10000000 // 10 megabytes max outstanding request size per SSE stream
+  private var droppedMessageCount: Long = 0 // in messages
+  private val maxBufferLength: Long = 100000 // 100k allowed to buffer before we send it.
 
   private val instanceId = NetflixEnvironment.instanceId()
 
-  private val tickTime = 30.seconds
+  private val tickTime = 100.microseconds
   private var tickCount = 0
+  private val statsTime = 300 // in 100ms increments.
   private val helloMessage = SSEHello(sseId, instanceId, GlobalUUID.get)
+  private var toSend = ""
 
-  client ! ChunkedResponseStart(HttpResponse(StatusCodes.OK)).withAck(Ack)
-  outstandingCount += 1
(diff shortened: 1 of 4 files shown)


atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
```diff
@@ -37,47 +37,56 @@ class SSEActor(client: ActorRef,
 
   private val connectsId = registry.createId("atlas.lwcapi.sse.connectCount")
   private val messagesId = registry.createId("atlas.lwcapi.sse.messageCount")
-  private val droppedId = registry.createId("atlas.lwcapi.sse.droppedCount")
+  private val messageBytesId = registry.createId("atlas.lwcapi.sse.messageBytesCount")
+  private val sendsId = registry.createId("atlas.lwcapi.sse.httpSendsCount")
+  private val sendBytesId = registry.createId("atlas.lwcapi.sse.httpSendBytesCount")
```
brharrington: I don't see where `sendsId` or `sendBytesId` are actually used.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
```diff
@@ -37,47 +37,56 @@ class SSEActor(client: ActorRef,
...
+  private val maxBufferLength: Long = 100000 // 100k allowed to buffer before we send it.
 
   private val instanceId = NetflixEnvironment.instanceId()
 
-  private val tickTime = 30.seconds
+  private val tickTime = 100.microseconds
   private var tickCount = 0
+  private val statsTime = 300 // in 100ms increments.
   private val helloMessage = SSEHello(sseId, instanceId, GlobalUUID.get)
+  private var toSend = ""
```
brharrington: I think this should be a [StringBuilder](https://docs.oracle.com/javase/8/docs/api/java/lang/StringBuilder.html) instead of a String so it is cheaper to append.

I found the `toSend` name a little confusing when first reading the code. How about something like `messageBuffer`?

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
```diff
@@ -37,47 +37,56 @@ class SSEActor(client: ActorRef,
...
 
-  client ! ChunkedResponseStart(HttpResponse(StatusCodes.OK)).withAck(Ack)
-  outstandingCount += 1
+  client ! ChunkedResponseStart(HttpResponse(StatusCodes.OK)).withAck(Ack(1))
+  outstandingByteCount += 1 // ignore byte lengh here, just send it.
```
brharrington: Why wouldn't the byte count and size for the Ack be 0 here?

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
```diff
@@ -88,15 +97,35 @@ class SSEActor(client: ActorRef,
...
+      queueToBuffer(msg)
+      flushBuffer()
     } else {
-      droppedCount += 1
-      registry.counter(droppedId.withTag("action", msg.getWhat)).increment()
+      if (outstandingByteCount < maxOutstandingBytes) {
+        queueToBuffer(msg)
+      } else {
+        droppedMessageCount += 1
+        registry.counter(droppedId.withTag("action", msg.getWhat)).increment()
```
brharrington: Should we add some sort of message so that the consumer knows they are not keeping up and that data is being dropped? The counter helps us when operating the service, but an individual consumer wouldn't know that they have partial data.

skandragon: We include droppedMessageCount in the periodic stats message, so I believe they should get that count to let them know the number of lost messages.  I felt lost bytes was less important here, so removed that.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
skandragon: Fixed.

atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/SSEActor.scala
skandragon: Fixed.

PR #510, opened 2017-02-23
allow `last` summary stat to be used with filters

brharrington:
The last value for a given time series is already being
tracked as part of the summary stats, but was not supported
by the `:stat` operator. This change allows that stat to
be used as part of a filter.


pparth:
This is not available in the latest 1.5.X versions, correct? Do we need to wait for 1.6.0 for this?

brharrington:
Correct. It is a small enough change that I suppose we could backport if there is demand.

You could also use one of the 1.6 RCs as long as you are comfortable with the caveats mentioned in:

https://github.com/Netflix/atlas/blob/master/CONTRIBUTING.md#versions-and-compatibility

PR #525, opened 2017-03-24
support inline rollup of datapoints

brharrington:
This change allows datapoints to be rolled up as they
are flowing into the in memory storage. Typically this
has been done later as part of offline processing using
EMR or via a separate aggregator cluster that would
perform the aggregations prior to being sent to the
in memory storage.

To be eligible for the rollup the data must have already
been normalized to the step boundary and rate conversion
for counters should have already been performed.

The inline rollup will create a rollup block so it will
have more overhead in terms of memory use unless the
reduction multiple is over 4. Data will only go through
the rollup path if it is tagged with `atlas.rollup=true`.

The publish payload can now take an `id` field which will
get used to dedup messages. If used, then it should avoid
over counting if retries are present in the publishing
pipeline as long as the updates for a given datapoint are
always mapped to the same id for a given interval. The
message id should not be used for multiple intervals or
subsequent messages would incorrectly get lost.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/db/BlockStore.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/db/BlockStore.scala
@@ -43,7 +43,8 @@ object BlockStats {
       case b: ArrayBlock    => inc(block, arrayCount, arrayBytes)
       case b: ConstantBlock => inc(block, constantCount, constantBytes)
       case b: SparseBlock   => inc(block, sparseCount, sparseBytes)
-      case _ =>
+      case b: RollupBlock   => b.blocks.foreach(inc)
+      case _                =>
     }
   }
 
@@ -57,7 +58,8 @@ object BlockStats {
       case b: ArrayBlock    => dec(block, arrayCount, arrayBytes)
       case b: ConstantBlock => dec(block, constantCount, constantBytes)
       case b: SparseBlock   => dec(block, sparseCount, sparseBytes)
-      case _ =>
+      case b: RollupBlock   => b.blocks.foreach(dec)
+      case _                =>
     }
   }
 
@@ -97,7 +99,7 @@ trait BlockStore {
   def cleanup(cutoff: Long): Unit
 
   /** Updates the store with a datapoint. */
-  def update(timestamp: Long, value: Double): Unit
+  def update(timestamp: Long, value: Double, rollup: Boolean = false): Unit
 
   /**
     * Force and update with no data. This can be used to trigger the rotation and compression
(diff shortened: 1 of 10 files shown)


pparth:
This looks promising. Do we have any documentation about it?

brharrington:
We're still testing it out and do not have docs yet. Internally we do this already with a separate aggregation cluster and the plan is to git rid of that separate cluster. The primary use-case for us is dropping the node dimension so we only retain the cluster level aggregate. So suppose we have 100 nodes each sending data like:

```json
{
  "tags": {
    "nf.cluster": "testapp-dev",
    "nf.node": "i-12345"
  } ,
  "metrics": [
    {
      "tags": {"name": "counterToRollup"},
      "timestamp": 1490630580000,
      "value": 42.0
    },
    {
      "tags": {"name": "counterNeededPerNode"},
      "timestamp": 1490630580000,
      "value": 42.0
    }
  ]
}
```

Can be rewritten to:

```json
{
  "tags": {
    "nf.cluster": "testapp-dev"
  } ,
  "metrics": [
    {
      "tags": {"name": "counterToRollup", "atlas.rollup": "true"},
      "timestamp": 1490630580000,
      "value": 42.0
    },
    {
      "tags": {"name": "counterNeededPerNode", "nf.node": "i-12345"},
      "timestamp": 1490630580000,
      "value": 42.0
    }
  ]
}
```

This means the `counterToRollup` would only have a single copy in memory, but would get mapped to a rollup block type that keeps track of the main stats. So the `:min`, `:max`, `:count`, and `:sum` would be the same as if it was not rolled up. The `counterNeededPerNode` would still have one copy per distinct node. I'll be updating the spectator AtlasRegistry to support either this week or next depending on how some other projects go this week.

pparth:
So, this will automatically calculate and keep values for the 4 rollup functions, for the specific metric? 
Is the grouping tag going to be configurable?

brharrington:
> So, this will automatically calculate and keep values for the 4 rollup functions, for the specific metric?

Yes

> Is the grouping tag name configurable?

There isn't really a grouping tag name. A metric is identified by the full set of tags, so if the tag set is the same and it is indicated as a rollup, then it will be aggregated instead of treated as another measurement of the same time series.

pparth:
Ah yes. I forget that a metric is essentially a specific tag set. So, the atlas.rollup tag name with a value of True, essentially marks the whole tag set as a rollup. So, whenever the same tag set is published to the server, the server will perform the rollup functionality for it. In your example, the rollup tag set is: "nf.cluster": "testapp-dev" , "name": "counterToRollup". Whenever i publish such a set, apparently from different nodes, if i do not include the rollup tag and not include the node tag, then data points from different nodes would overlap in a single, confusing time series. If i include the rollup tag and remove the node tag, then 4 aggregated time series would be maintained.

PR #526, opened 2017-03-25
improve performance of SmallHashMap.{+,-}

brharrington:
Changes the small hash map implementation of `+`
and `-` to retain the type. This is primarily to
avoid using a less memory efficient map type if
the map is modified somewhere along the way. A
side benefit is better performance and reduced
allocations.

**Before**

```
Benchmark        Mode  Cnt       Score       Error  Units
addPair         thrpt   10  781344.330 ± 25173.790  ops/s
removePair      thrpt   10  738796.252 ± 44849.997  ops/s
```

**After**

```
Benchmark        Mode  Cnt        Score       Error  Units
addPair         thrpt   10  1637470.772 ± 29948.315  ops/s
removePair      thrpt   10  2520939.662 ± 32320.759  ops/s
```

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/util/SmallHashMap.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/util/SmallHashMap.scala
@@ -16,18 +16,18 @@
 package com.netflix.atlas.core.util
 
 object SmallHashMap {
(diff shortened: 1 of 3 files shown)


brharrington:
**Allocations Before**

screen shot 2017-03-25 at 12 11 45 pm

**Allocations After**

screen shot 2017-03-25 at 12 16 39 pm

PR #535, opened 2017-04-11
only run lwc when using the lwcapi.conf

brharrington:
Fixes the standalone example to only run the lwc api
and services when using the lwcapi config. The webapi
portions have been restricted in a similar way. This
avoids the need for having redis to run the basic
standalone example.

### DIFF
--- a/atlas-standalone/src/main/resources/application.conf
+++ b/atlas-standalone/src/main/resources/application.conf
@@ -1,0 +1,20 @@
+atlas {
+  akka {
+    // These are overridden because the standalone pulls in both webapi and lwc modules.
+    // The user can run the one that is needed for a given test by using one of the specific
+    // profile configurations. The default application config here ignores the setup done
+    // by the reference confs.
+    actors = [
+      {
+        name = "deadLetterStats"
+        class = "com.netflix.atlas.akka.DeadLetterStatsActor"
+      }
+    ]
+
+    api-endpoints = [
+      "com.netflix.atlas.akka.HealthcheckApi",
+      "com.netflix.atlas.akka.ConfigApi",
+      "com.netflix.atlas.akka.StaticPages"
+    ]
+  }
+}
--- a/atlas-standalone/src/main/resources/static.conf
+++ b/atlas-standalone/src/main/resources/static.conf
@@ -1,0 +1,17 @@
+
+atlas {
+  akka {
+    actors = ${?atlas.akka.actors} [
+      {
+        name = "db"
+        class = "com.netflix.atlas.webapi.LocalDatabaseActor"
+      }
+    ]
+
+    api-endpoints = ${?atlas.akka.api-endpoints} [
+      "com.netflix.atlas.webapi.TagsApi",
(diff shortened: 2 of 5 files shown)


pparth:
@brharrington  Can you please provide more info on what lwc api is about?

brharrington:
The name may change it came from the internal project: light-weight client (LWC). Historically the internal java client used for Atlas has quite a bit of functionality including:

* [On-instance alerting](https://github.com/Netflix/atlas/wiki/Alerting-Philosophy#alerting-scopes)
* Publishing custom metrics to CloudWatch for auto-scaling
* Local filtering and rollups to help manage capacity
* Redirect do inline aggregator cluster for cross-instance rollups
* Admin interface for debugging
* Streaming hook to allow [Mantis](http://techblog.netflix.com/2016/03/stream-processing-with-mantis.html) access the data

There are a number of internal trends motivating us to thin down the client while retaining as much of the functionality as possible. In particular:

* Pace of adoption. Changes to the client take a long time to propagate and get deployed to all applications.
* Polyglot support. Java was the only supported language for a long time and we are also supporting node.js now. The more capability we have in the client the more painful the duplication becomes.
* Streaming use-cases are becoming more prevalent and the current pull model causes a number of issues. This is also expected to be an option for improved capacity savings in the future because we may have some data that is only available on-demand for stream processing and not stored all the time.
* Cleanup and reduce user confusion. Many of these use-cases came up organically over the years and the set of client side options the user has to comprehend are quite large.

Many of these use-cases follow the same pattern of specifying an expression that evaluates locally over the data and sends it somewhere. For example:

* _On-instance alerts_: you configure a set of [alerting expressions](https://github.com/Netflix/atlas/wiki/Alerting-Expressions), they are evaluated each interval, and if in a triggered state notify the internal alerting api.
* _CloudWatch publishing_: you configure an expression similar to a graph and it sends that as a metric to CloudWatch that can be used in an auto-scaling rule.
* _Streaming hook_: listens on admin service and you can subscribe with a query expression to get an SSE stream with the data for that query.

LWC API is providing a way for various tools to subscribe for a given expression and get the data. The client just needs to know how to communicate with LWC and that only requires the [query](https://github.com/Netflix/atlas/wiki/Reference-query) and [data](https://github.com/Netflix/atlas/wiki/Reference-data) expression support. Those operations are relatively stable and keeping them on the client means we can get a big data reduction up front. The other use-cases would be maintained off instance using this API.

The downside to this is for use-cases like on-instance alerts and CloudWatch publishing one of the main original goals of running on-instance was to minimize the amount of infrastructure needed for them to work. So we'll have a higher probability of failure for those use-cases, but overall we feel those risks can be mitigated to an acceptable level. 

The other change for stream processing in particular is that the instance will always push the data and connects out to the LWC service to get the subscriptions rather than require something to connect into the instance. This is generally preferred because with increasing adoption of faster infrastructure such as containers instances can come up and be available before external services learn about them and try to connect in. So this helps avoid startup gaps and only allowing incoming traffic for the monitoring. It is also an easier sell for some of our more secure environments where we need to be very careful about incoming connectivity.

PR #548, opened 2017-04-24
simplify eval usage for common use-cases

### LINKED ISSUE #539: support dynamic set of urls for streams
In some cases it is desirable to dynamically choose the set of hosts and reuse connections for sending the data. The flow would look something like:

```
({add,remove}, id, url) > evaluate-urls > (id, data)
```

brharrington:
Fixes #539. 

Adds an `Evaluator` class with a simple public API
for evaluating the stream. It is a java class that
exposes the reactive streams types so it is easy to
use with various implementations. There are two
main methods:

**createPublisher(uri)**

Creates a publisher for a graph uri so the same data
can be accessed as a stream. This makes it really easy
for a user to transition from one view to another.

```
+------------+
|  URI       |
|  Eval      | -------- [TimeSeriesMessage] --->
|  Source    |
+------------+
```

**createStreamsProcessor**

Takes a set of id/URI pairs as inputs and evaluates
all of the expressions. This allows for better reuse
when evaluating many expressions. There is an id
for each input that will be added to the output so
that corresponding entries can be correlated.
The complete set of current data sources must be
supplied as part of each input event.

```
                       +------+
                       |      |
--- [DataSources] ---> | Eval | --- [MessageEnvelope] --->
  Set((id,URI), ...)   |      |   (id, TimeSeriesMessage)
                       +------+
```


atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaSource.scala
```diff
@@ -0,0 +1,204 @@
...
+    *     Eureka service. Depending on the Eureka service configuration there may be
+    *     some variation in the path.
+    * @param instanceUriPattern
+    *     Pattern for constructing the URI for each instance in Eureka. The allowed
+    *     substitutions are `{port}` and any key in the data center info metadata
+    *     block for an instance. Example:
+    *
+    *     ```
+    *     http://{local-ipv4}:{port}/metrics?q=name,cpu,:eq,:sum
+    *     ```
```
brharrington: Client param missing.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
```diff
@@ -0,0 +1,207 @@
...
+  case class EurekaBackend(host: String, eurekaUri: String, instanceUri: String) extends Backend {
+    def source: Source[ByteString, _] = EurekaSource(eurekaUri, instanceUri, null)
+
+    def run[T](sink: Sink[ByteString, T])
+      (implicit system: ActorSystem, materializer: ActorMaterializer): StreamRef[T] = {
+
+      val client = Http().superPool[NotUsed]()
+      val src = EurekaSource(eurekaUri, instanceUri, client)
+      EvaluationFlows.run(src, sink)
+      //EvaluationFlows.runForEurekaVip(eurekaUri, instanceUri, sink)
```
brharrington: Remove commented out code

PR #551, opened 2017-04-25
add eval lib to root project

brharrington:
It wasn't getting published to jfrog/jcenter
since it wasn't part of the root.


brharrington:
Tests were not running before as part of validation build, needs #550 for tests to pass.

brharrington:
Closing in favor of #553.

PR #592, opened 2017-05-19
illustrate how to construct alerts for low volume conditions

copperlight:


copperlight:
Looks good, thanks for the edits!

PR #635, opened 2017-07-26
Cross reference data/:max page in math/:max and vice versa

narayanan-arunachalam:


brharrington:
I published the wiki, changes are live.

PR #683, opened 2017-09-18
[atlas-eval] Changing logLevel for instances added/removed from info …

ruchirj:


brharrington:
Note, in the future, please use a private fork and submitting the PR from the fork rather than pushing a branch to the main repo.

PR #687, opened 2017-09-19
improve performance for bulk subscribe

brharrington:
When creating a subscription with many expressions, the
`updateSubHandlers()` method in the subscription manager
is quite expensive. Before, the subscribe api would end
up calling this once per expression, now it will only
call it once.

/cc @ruchirj


brharrington:
screen shot 2017-09-19 at 10 48 56 am

ruchirj:
Looks good.

PR #699, opened 2017-10-06
remove handlers when a stream is unregistered

brharrington:
Before when the stream went away the map for the subscription
to the handler was not getting cleaned up.

### DIFF
--- a/atlas-lwcapi/src/test/scala/com/netflix/atlas/lwcapi/SubscriptionManagerSuite.scala
+++ b/atlas-lwcapi/src/test/scala/com/netflix/atlas/lwcapi/SubscriptionManagerSuite.scala
@@ -135,9 +135,9 @@ class SubscriptionManagerSuite extends FunSuite {
 
     val s = sub("name,exp1,:eq")
     sm.subscribe("a", s)
-    assert(sm.handlersForSubscription(s.metadata.id).sorted === List(1))
+    assert(sm.handlersForSubscription(s.metadata.id) === List(1))
 
     sm.unregister("a")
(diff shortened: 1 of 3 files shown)


atlas-lwcapi/src/test/scala/com/netflix/atlas/lwcapi/SubscriptionManagerSuite.scala
```diff
@@ -128,4 +128,16 @@ class SubscriptionManagerSuite extends FunSuite {
...
     assert(sm.unregister("a") === None)
   }
+
+  test("unregister should remove handlers") {
+    val sm = new SubscriptionManager[Integer]()
+    sm.register("a", 1)
+
+    val s = sub("name,exp1,:eq")
+    sm.subscribe("a", s)
+    assert(sm.handlersForSubscription(s.metadata.id).sorted === List(1))
```
zimmermatt: Minor nit.. the .sorted is unnecessary.

atlas-lwcapi/src/test/scala/com/netflix/atlas/lwcapi/SubscriptionManagerSuite.scala
```diff
@@ -128,4 +128,16 @@ class SubscriptionManagerSuite extends FunSuite {
...
+
+    sm.unregister("a")
+    assert(sm.handlersForSubscription(s.metadata.id).sorted === Nil)
```
zimmermatt: Minor nit.. the .sorted is unnecessary.

PR #703, opened 2017-10-09
consume Eureka response body if status not OK

brharrington:
Fixes the code to discard the entity when the status
is something other than 200 OK.

### DIFF
--- a/atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/EurekaSourceSuite.scala
+++ b/atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/EurekaSourceSuite.scala
@@ -17,6 +17,8 @@ package com.netflix.atlas.eval.stream
 
 import java.io.IOException
 import java.nio.charset.StandardCharsets
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.HttpEntity
@@ -32,8 +34,10 @@ import akka.stream.scaladsl.Flow
 import akka.stream.scaladsl.Sink
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
+import com.netflix.atlas.akka.AccessLogger
 import com.netflix.atlas.core.util.Streams
 import com.netflix.atlas.eval.stream.EurekaSource.GroupResponse
+import com.typesafe.config.ConfigFactory
 import org.scalatest.FunSuite
 
 import scala.concurrent.Await
@@ -94,9 +98,14 @@ class EurekaSourceSuite extends FunSuite {
   private implicit val system = ActorSystem(getClass.getSimpleName)
   private implicit val mat = ActorMaterializer()
 
+  private val config = ConfigFactory.parseString("backends = []")
+
   private def run(uri: String, response: Try[HttpResponse]): GroupResponse = {
-    val client = Flow[HttpRequest].map(_ => response)
-    val future = EurekaSource(uri, client).runWith(Sink.head)
+    val client = Flow[(HttpRequest, AccessLogger)].map {
+      case (_, logger) => response -> logger
+    }
+    val context = new StreamContext(config, client, mat)
+    val future = EurekaSource(uri, context).runWith(Sink.head)
     Await.result(future, Duration.Inf)
   }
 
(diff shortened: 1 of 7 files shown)


atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/EurekaSourceSuite.scala
```diff
@@ -138,6 +146,23 @@ class EurekaSourceSuite extends FunSuite {
     }
   }
```
zimmermatt: I think this can be simplified to:

```
  test("entity for bad status code is consumed") {
    var consumed = false
    val source = Source
      .single(ByteString.empty)
      .map { data =>
        consumed = true
        data
      }
    val entity = HttpEntity(MediaTypes.`application/json`, source)
    val response = HttpResponse(StatusCodes.BadRequest, entity = entity)
    val uri = "http://eureka/v1/vips/www-dev:7001"
    intercept[NoSuchElementException] {
      run(uri, Success(response))
    }
    assert(consumed)
  }
```

The `await` will hang in the failure case and there doesn't seem any need for coordination.  However, another option to avoid the hang, but may introduce fragility, would be to change:
```
    consumedLatch.await()
```
to
```
    assert(consumedLatch.await(3, TimeUnit.SECONDS))
```

brharrington: The reason for the coordination is that there is no guarantee that the entity will be discarded before the exception is thrown. It can be processed asynchronously from the main flow.

Personally I would rather not have an arbitrary timeout, though I suppose we could just set it high. Why are you worried about it blocking in the failure case?

zimmermatt: If it blocks indefinitely, It tends to be more disruptive to get out of (i.e., you end up having to terminate the overall build).  If it fails from a time out, the other tests can run and the overall build will complete without intervention.

brharrington: The build will complete anyway, travis will timeout after 10m of no output. The failure will need to be addressed so I'm not too concerned about the other tests running in the case it blocks. My main concern is the source of the problem is clear and it is not flakey. The source of the problem should be clear either way, any sort of short timeout is likely to be flakey on a stressed build machine. As a compromise I used a timeout of 30s.

zimmermatt: Thanks for implementing with a compromise.

For posterity, I'll comment here with a more context why I'm pushing this, given that this is likely to be neither a high probability nor common occurrence.

At a fundamental level our goals are consistent in regard to stability and clarity.  I'm thinking about the local build case.  I often will kick off a build, either in the IDE or at the command line, and then switch to something else (email or some other small task).  I would feel frustrated if I then switched back and found it hanging.

In regard to clarity, the local command line build (and quite possibly in travis) is not informative if it hangs.  I ran a test with the `CountDownLatch::countDown()` commented out.  All of the tests did run.  However, neither `EurekaSourceSuite` nor the test description (`entity for bad status code is consumed`) was referenced anywhere in the output.  In addition, the console didn't return to the prompt and sat forever on the last line in the following:

```
[info] - /api/v1/graph?q=42&s=e-30y&e=2017-06-27T00:00&tz=UTC&w=430&h=120&layout=iw
[info] - /api/v1/graph?s=e-2d&e=2012-01-01T00:00&q=name,sps,:eq,(,nf.node,),:by
[info] Run completed in 1 minute, 22 seconds.
[info] Total number of tests run: 246
[info] Suites: completed 10, aborted 0
[info] Tests: succeeded 246, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
```

When I pressed `Ctrl-C` to kill it, nothing informative was printed; only the following:

```
[info] All tests passed.
^CSLF4J: Failed to load class "org.slf4j.impl.StaticMDCBinder".
SLF4J: Defaulting to no-operation MDCAdapter implementation.
SLF4J: See http://www.slf4j.org/codes.html#no_static_mdc_binder for further details.
SLF4J: Failed to load class "org.slf4j.impl.StaticMDCBinder".
SLF4J: Defaulting to no-operation MDCAdapter implementation.
SLF4J: See http://www.slf4j.org/codes.html#no_static_mdc_binder for further details.
SLF4J: Failed to load class "org.slf4j.impl.StaticMDCBinder".
SLF4J: Defaulting to no-operation MDCAdapter implementation.
SLF4J: See http://www.slf4j.org/codes.html#no_static_mdc_binder for further details.
make: *** [build] Error 130
```

With a sufficiently large timeout, it seems we get the best of both dimensions.

Issue #694, opened 2017-09-28
improve diagnostic messages for lwc evaluation


brharrington:
Update the diagnostic messages to provide more detail about the amount of data for an expression and time window. This can be helpful for the end user when debugging and avoids them just seeing nothing for a given expression.

PR #723, opened 2017-10-26
improve diagnostic messages for lwc evaluation (#694).

zimmermatt:
Insert diagnostic message summarizing input datapoint counts and
post aggregation counts for time series expressions.

### DIFF
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
@@ -15,6 +15,8 @@
  */
 package com.netflix.atlas.eval.stream
 
+import java.time.Instant
+
 import akka.NotUsed
 import akka.http.scaladsl.model.Uri
 import akka.stream.Attributes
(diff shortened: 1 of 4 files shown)


atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
```diff
@@ -122,26 +122,44 @@ private[stream] class FinalExprEval(step: Long = 60000L)
...
+              val result = expr.expr.eval(context, expressionDatapoints)
               states(expr) = result.state
               val msgs = result.data.map { t =>
                 TimeSeriesMessage(expr.toString, context, t)
               }
+
+              val diagnostics = expr.expr.dataExprs.flatMap(expressionDiagnostics.get)
+
               ids.flatMap { id =>
                 msgs.map { msg =>
```
brharrington: `(msgs ++ diagnostics).map`

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
```diff
@@ -122,26 +122,44 @@ private[stream] class FinalExprEval(step: Long = 60000L)
...
+                s"Input of ${datapoints.length} datapoints further aggregated to ${timeseries.length} datapoints.")
+
+              dataBldr += dataExpr -> timeseries
+              diagBldr += dataExpr -> diagnosticMessage
+
+              (dataBldr, diagBldr)
+          }
+
+        val expressionDatapoints = noData ++ exprDataBldr.result()
+        val expressionDiagnostics = exprDiagBldr.result()
```
brharrington: I think it would be easier to read like:

```scala
val groupedDatapoints = group.values.groupBy(_.expr)
val data = noData ++ groupedDatapoints.map {
  case (k, vs) =>
    k -> AggrDatapoint.aggregate(vs).map(_.toTimeSeries(step))
}
val diagnostics = groupedDatapoints.map {
  case (k, vs) =>
    val t = Instant.ofEpochMilli(group.timestamp)
    k -> DiagnosticMessage.info(s"$t: ${vs.length} input datapoints for [$k]")
}
```

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
zimmermatt: Good suggestion. Will do.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
zimmermatt: I had initially done a version like this, but wanted to keep it to one traversal. I'm not tied to the fold though.

zimmermatt: Oh... and I also wanted to get the length of the collection returned from `AggrDatapoint.aggregate(datapoints)` in the `DiagnosticMessage`.

brharrington: Summarizing in person discussion:

I don't think the output size is all that useful. In most cases the users can use just look at the data being emitted and in particular when reducing an expression for debugging I usually restrict it to just a data expression. So the full output is available to the user.

PR #731, opened 2017-10-28
update to sbt 1.0.3 and scala 2.12.4

brharrington:
SBT 1.0.3 fixes the problem we were seeing in #725. Tried
reimporting and I can still run tests with no issues.


brharrington:
Build is failing on jdk9 due to: https://github.com/sbt/sbt/issues/3684

PR #736, opened 2017-11-11
fix #733, add :freeze operator

### LINKED ISSUE #733: make it easier to create overlay graphs
When appending expressions that use the `:list,(,[refinement query],:cq,),:each` it becomes difficult to combine them for the purposes of creating an overlay. Ideally there would be some way to indicate that they should have isolated stacks.

/cc @vfilanovsky

brharrington:
Adds `:freeze` operator to allow freezing a portion
of the stack so it is isolated from further modification.
This is useful for query overlay and some other use-cases
where we essentially need to combine multiple expressions
and still be able to use operations like `:clear` or `:list`
without worrying about exactly how far back to go on the
stack.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/stacklang/Context.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/stacklang/Context.scala
@@ -15,4 +15,40 @@
  */
 package com.netflix.atlas.core.stacklang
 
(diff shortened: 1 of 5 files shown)


pparth:
Can you provide please some documentation about query overlaying using Atlas?

brharrington:
There are some docs on the page for the operator:

https://github.com/Netflix/atlas/wiki/std-freeze

PR #752, opened 2017-12-12
use the decoded name as field alias

brharrington:
The name can have encoded characters such as `.`. For
example `a.b` will be encoded as `a$u002Eb`. The decoded
name is now used as an alias if there is not an explict
alias set with an annotation.


copperlight:
This works great, thanks for the quick turn-around!

PR #759, opened 2018-01-05
add mapping for TrustedAdvisor metrics

brharrington:
For now it is only the service limit usage metrics that
give us an idea of how close we are to limits. This can
be a simple alternative to many of the uses of howlermonkey.
The update frequency for the trusted advisor metrics is
fairly low so it will take an hour or so to detect a
problem. For slow growth it is adequate.


pparth:
Is there any documentation regarding the ability to have Atlas server poll and add CloudWatch metrics?

brharrington:
Not currently, I filed #760 to track.

PR #770, opened 2018-01-30
Provide a more useful description for `:dist-avg` and `:dist-max`.

zimmermatt:
The generated defaults are not very helpful to users since they just
output the expanded rewrite syntax.


zimmermatt:
The dist-avg looks good.  The dist-max is a little hard to parse, IMO. I'll propose an alternate.

PR #797, opened 2018-02-27
initial work to split rendering from webapi

brharrington:
There are a number of use-cases where it is convenient to
be able to render a chart based on the uri without going
through the webapi. Examples:

- Alerting through LWC with a certain buffer for the data
  being streamed through. If an alert triggers, then it
  is desirable to be able to include context in the email
  or other notification. The graph image helps the user
  understand why it fired.
- Spark queries over the data in S3. We want to be able
  to let a user use the exact same graph uri they use for
  the main backend and run it on spark.

This is an initial step and we'll try it out a bit to see
what changes make sense. A subset of the tests from the
GraphApiSuite have been moved over. Most of the rest will
follow later on so we can ensure that there isn't anything
being added in the graph layer.


brharrington:
Verified that chart generation for wiki still works as expected.

PR #811, opened 2018-04-02
update dependencies

brharrington:


brharrington:
There seems to be a problem with the pom.sha1 file for akka-http_2.11 for 10.1.1 in jcenter (akka/akka-http#1972). Can probably work around by adding a maven central resolver.

PR #812, opened 2018-04-03
reduce number of nodes in demo dataset

brharrington:
This is used for test cases and docs. Reducing the number
of nodes significantly improves the time for running tests
or generating docs that use longer time ranges.


brharrington:
Seems to help more locally than it does on travis.

PR #865, opened 2018-08-03
Add frequency parameter to `DataSource`.

zimmermatt:
As part of enabling high resolution metrics for the streaming path, this
commit adds a frequency parameter to `DataSource`. This is then used to
configure the frequency in the metrics subscription.

I tested this by running an lwcapi server locally and observing that the
frequency listed in the response from `/api/v1/expressions` matched the
overridden value.

Note that this is an incremental step. For this to fully function, the
frequency needs to be introduced into the `Flow` that groups and
evaluates the datapoints. This will be done in a following commit.


atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
```diff
@@ -181,11 +181,41 @@ DataSources localOnly() {
   public final static class DataSource {
     private final String id;
     private final String uri;
-
-    /** Create a new instance. */
+    private final Duration frequency;
+
+    /** The default frequency of metrics publication. */
+    public static final Duration DEFAULT_FREQUENCY = Duration.ofSeconds(60L);
```
brharrington: This should probably be private for now. Longer term the default frequency should go away or come from the configuration.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
```diff
@@ -181,11 +181,41 @@ DataSources localOnly() {
...
+
+    /**
+     *  Create a new instance with the {@link #DEFAULT_FREQUENCY}.
+     *
+     * @param id
+     *     An identifier for this {@code DataSource}.
```
brharrington: Consider something like:

> Id that will be added to corresponding MessageEnvelope objects in the output. This allows the user to match output messages with the data source.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
```diff
@@ -181,11 +181,41 @@ DataSources localOnly() {
...
+     *  Create a new instance.
+     *
+     * @param id
+     *     An identifier for this {@code DataSource}.
+     * @param uri
+     *     The URI for this {@code DataSource} (in atlas backend form).
+     * @param frequency
+     *     The requested frequency of metrics publication for this
+     *     {@code DataSource} in atlas backend form, with the ASL query as a
+     *     {@code query} param. <em>NOTE:</em> This may result in rejection of
```
brharrington: Why is `query` a code span?

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
```diff
@@ -181,11 +181,41 @@ DataSources localOnly() {
...
+    /**
+     *  Create a new instance.
+     *
+     * @param id
+     *     An identifier for this {@code DataSource}.
+     * @param uri
+     *     The URI for this {@code DataSource} (in atlas backend form).
+     * @param frequency
+     *     The requested frequency of metrics publication for this
+     *     {@code DataSource} in atlas backend form, with the ASL query as a
```
brharrington: s/query/expression/

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
```diff
@@ -181,11 +181,41 @@ DataSources localOnly() {
...
+     *     {@code query} param. <em>NOTE:</em> This may result in rejection of
+     *     this {@code DataSource} if the backing metrics producers do not
+     *     support the requested frequency.
+     */
+    public DataSource(String id, String uri, Duration frequency) {
       this.id = id;
       this.uri = uri;
+      this.frequency = frequency;
```
brharrington: Frequency should be included in the `equals` and `hashCode` check. I have some added test cases that would catch this in #866.

zimmermatt:
Thanks. I was hoping I was going to finish a follow-on before you got to this. I realized I missed `equals`/`hashCode`, and `toString` as well. An errant Cmd+V seems to have messed up one of the doc comments.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
zimmermatt: ok

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
zimmermatt: ok

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
zimmermatt: An errant Cmd+V in the middle seems to have messed this up (it's from an earlier version of the URI doc).

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
zimmermatt: Yeah... annoyed at myself I missed this.

zimmermatt:
I've been trying to add JSON serialization support for the use of `java.time.Duration` by adding https://mvnrepository.com/artifact/com.fasterxml.jackson.module/jackson-modules-java8 to atlas-json. However even though the ivy metadata files show up in my `~/.ivy2/cache`, and I've added it to  the Dependencies.scala file and to the atlas-json dependency list, it's not building (can't find `JavaTimeModule` when I try to register it). I even nuked my iv2 cache to pull fresh.

I'll try again tomorrow. If it doesn't work then, I'll either move to JodaTime or change the `frequency` parameter to a primitive.

brharrington:
Joda is mostly phased out for Atlas. It is only kept in the json library because of some other apps that pull in `atlas-json`. I'll probably drop joda entirely in 1.7.x and users can add the module themselves if they still need it.

I tried the java8 modules and it worked fine for me (#867) though I wish the serialized representation was the same as the joda variants. I'm guessing they had to change it to avoid losing precision though.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
```diff
@@ -213,17 +254,17 @@ public boolean isLocal() {
...
     @Override public int hashCode() {
       int result = id.hashCode();
-      result = 31 * result + uri.hashCode();
+      result = 31 * result + uri.hashCode() + frequency.hashCode();
       return result;
     }
 
     @Override public String toString() {
-      return "DataSource(" + id + "," + uri + ")";
+      return "DataSource(" + id + "," + frequency + "," + uri + ")";
```
brharrington: Ideally this would match the positions of the parameters in the constructor to avoid confusion.

zimmermatt:
For context, I put the `frequency` before the URI in `toString` for readability.  Since `,` is used in ASL expressions, it will visually look like it's part of the URI if placed after. This will likely result in extra effort to understand what one is seeing when printed.

Perhaps there's another way to keep it readable... maybe by changing the constructor order.  Though that then ends up being a little awkward because of the default value. We should consider removing the default `frequency` constructor soon than later if we go this route. Another option is changing the field delimiter.

PR #866, opened 2018-08-03
add test cases for Evaluator model objects

brharrington:
Verify the equals/hashCode contracts are satisfied. Also
check that DataSource and DataSources can be encoded and
decoded from JSON correctly. Needed to consume for replay
logging.


brharrington:
Hmm, fails in sbt and passes in the IDE.

PR #887, opened 2018-09-03
Value based rule constructors

janstenpickle:
Allow rules to be constructed from values as well from Typesafe Config. This allows rules to be constructed by other means such as configuration stored in a database, while maintaining the original reflection based instantiation.

This changes the implementation for the reflection based rule instantiation, but does not change the configuration format (as can be seen in the test). I have also changed the `Rule` implementations to be case classes so that they are serializable (which has caused me issues with Spark in the past) and can be instantiated via shapeless' `Generic`/`LabelledGeneric` (also useful to me!).


atlas-core/src/main/scala/com/netflix/atlas/core/validation/Rule.scala
```diff
@@ -58,7 +60,22 @@ object Rule {
   def load(ruleConfigs: List[_ <: Config]): List[Rule] = {
     ruleConfigs.map { cfg =>
       val cls = Class.forName(cfg.getString("class"))
-      cls.getConstructor(classOf[Config]).newInstance(cfg).asInstanceOf[Rule]
+
+      val runtimeMirror = universe.runtimeMirror(cls.getClassLoader)
+      val moduleSymbol = runtimeMirror.moduleSymbol(cls)
```
brharrington: If I understand correctly, then this would require the rules to have a companion object that takes the config via an apply method. Any existing rules that are not part of this repo would potentially break. Need to evaluate how much of a concern that would be for us. Also, this would make it more difficult to have a Java implementation of a rule. This should probably fallback to looking for a matching constructor.

atlas-core/src/main/scala/com/netflix/atlas/core/validation/Rule.scala
```diff
@@ -58,7 +60,22 @@ object Rule {
...
+
+      val targetMethod = moduleSymbol.typeSignature.members.collect {
+        case x
+            if x.isMethod && x.name.toString == "apply" && x.asMethod.paramLists.flatten.forall(
```
brharrington: Isn't the paramLists match overly broad? Wouldn't this match a method that took two Config parameters for example?

janstenpickle:
Thanks for looking at this so quickly! I've made a few updates based on your feedback.

atlas-core/src/main/scala/com/netflix/atlas/core/validation/ValuePatternRule.scala
```diff
@@ -26,9 +26,7 @@ import com.typesafe.config.Config
   * pattern = "^[-_.a-zA-Z0-9]{4,60}$"
   * ```
   */
-class ValuePatternRule(config: Config) extends TagRule {
-
-  private val pattern = Pattern.compile(config.getString("pattern"))
+class ValuePatternRule(pattern: Pattern) extends TagRule {
```
brharrington: Case class?

PR #895, opened 2018-09-08
extract the name portion of the target group dimension

zimmermatt:
Some dimension values are decorated with information that reduces
readability without providing value (e.g., ALBs and NLBs). This commit
introduces a new `extractors` configuration within the `tagger` that
enables specifying a regex to capture the relevant portion of the name.

### DIFF
--- a/atlas-poller-cloudwatch/src/main/resources/reference.conf
+++ b/atlas-poller-cloudwatch/src/main/resources/reference.conf
@@ -54,7 +54,7 @@ atlas {
       // Only one capture group is supported, so only capture group 1 will
       // be used. If no match is found, the unaltered original value is
       // used.
-      extractors = [
+      extractors = ${?atlas.cloudwatch.tagger.extractors} [
         {
           name = "TargetGroup"
           pattern = "^targetgroup/([^/]+)/.*"
--- a/atlas-poller-cloudwatch/src/main/resources/reference.conf
+++ b/atlas-poller-cloudwatch/src/main/resources/reference.conf
@@ -50,6 +50,17 @@ atlas {
     tagger = {
       class = "com.netflix.atlas.cloudwatch.NetflixTagger"
 
+      // Enables extracting a substring of a dimension value using a regex.
+      // Only one capture group is supported, so only capture group 1 will
+      // be used. If no match is found, the unaltered original value is
+      // used.
+      extractors = [
+        {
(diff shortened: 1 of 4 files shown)


atlas-poller-cloudwatch/src/main/resources/reference.conf
```diff
@@ -50,6 +50,17 @@ atlas {
...
       class = "com.netflix.atlas.cloudwatch.NetflixTagger"
 
+      // Enables extracting a substring of a dimension value using a regex.
+      // Only one capture group is supported, so only capture group 1 will
+      // be used. If no match is found, the unaltered original value is
+      // used.
+      extractors = [
+        {
+          name = "TargetGroup"
+          pattern = "^targetgroup/([^/]+)/.*"
```
brharrington: Should this go in `alb.conf` to improve locality of related configs?

atlas-poller-cloudwatch/src/main/resources/reference.conf
```diff
@@ -50,6 +50,17 @@ atlas {
     tagger = {
       class = "com.netflix.atlas.cloudwatch.NetflixTagger"
 
+      // Enables extracting a substring of a dimension value using a regex.
+      // Only one capture group is supported, so only capture group 1 will
+      // be used. If no match is found, the unaltered original value is
+      // used.
+      extractors = [
```
brharrington: Consider doing this with a conditional list append so that if extractors are put locally in the conf file for a given type they will append instead of clobbering each other.

atlas-poller-cloudwatch/src/main/resources/reference.conf
zimmermatt: Hmm... good question. I placed it here for locality to the `mappings`. I had originally thought to make it part of the `mappings`, but that would have introduced an awkwardness, since the need is infrequent.

But this introduces the question of whether the mappings for a particular component should also be moved to the related configs.

I'll leave it here for now. Once we have the NLB config in, we'll be able to see if these remain easily separable (i.e., no overlap). That will have the added benefit from a lineage standpoint of moving them around in a single separate commit.

atlas-poller-cloudwatch/src/main/resources/reference.conf
zimmermatt: will do

PR #898, opened 2018-09-11
add mappings for alb auth stats

zimmermatt:
This commit adds mappings for ApplicationELB authorization metrics
across two aliases.
* ELBAuthError, ELBAuthFailure, and ELBAuthSuccess are aliased to
  `aws.alb.authRequest` as a counter with a `result` tag with values of
  `error`, `failure`, and `success`.
* ELBAuthLatency is aliased to `aws.alb.authLatency` as a timer.

### DIFF
--- a/atlas-poller-cloudwatch/src/main/resources/alb.conf
+++ b/atlas-poller-cloudwatch/src/main/resources/alb.conf
@@ -46,11 +46,6 @@ atlas {
           ]
         },
         {
-          name = "ELBAuthLatency"
-          alias = "aws.alb.authLatency"
-          conversion = "timer"
-        },
-        {
           name = "ELBAuthSuccess"
           alias = "aws.alb.authRequest"
           conversion = "sum,rate"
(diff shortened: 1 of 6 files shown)


atlas-poller-cloudwatch/src/main/resources/alb.conf
```diff
@@ -23,6 +23,44 @@ atlas {
...
+          ]
+        },
+        {
+          name = "ELBAuthLatency"
+          alias = "aws.alb.authLatency"
+          conversion = "timer"
+        },
+        {
+          name = "ELBAuthSuccess"
+          alias = "aws.alb.authRequest"
```
brharrington: It would be nice to keep all of the values for `aws.alb.authRequest` together rather than have this one separated by the latency metric.

zimmermatt: Will do. I was thinking alphabetical by name, but keeping them together by alias makes sense.

PR #902, opened 2018-09-15
switch to spectator-ext-ipc for request metrics

brharrington:
Makes Atlas consistent with other Netflix apps following
the common-ipc pattern internally.

### DIFF
--- a/atlas-akka/src/main/scala/com/netflix/atlas/akka/AccessLogger.scala
+++ b/atlas-akka/src/main/scala/com/netflix/atlas/akka/AccessLogger.scala
@@ -17,7 +17,9 @@ package com.netflix.atlas.akka
 
 import akka.http.scaladsl.model.HttpRequest
 import akka.http.scaladsl.model.HttpResponse
-import com.netflix.spectator.sandbox.HttpLogEntry
+import com.netflix.spectator.api.Spectator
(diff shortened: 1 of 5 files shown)


atlas-akka/src/main/scala/com/netflix/atlas/akka/AccessLogger.scala
```diff
@@ -44,15 +46,15 @@ class AccessLogger private (entry: HttpLogEntry, client: Boolean) {
   def chunkStart(request: HttpRequest, response: HttpResponse): Unit = {
     AccessLogger.addRequestInfo(entry, request)
     AccessLogger.addResponseInfo(entry, response)
-    writeLog(AccessLogger.ChunkedStart)
+    entry.markEnd().log()
```
zimmermatt: `entry.markStart().log()`?
or was end intentional?

brharrington: It is intentional, it is started just before the request is sent out. By the time the first chunk arrives we already have the response, but may not have the complete request and the request may last a long time. For example, SSE responses could last for hours. In the case of chunked payloads, this logger is indicating the latency to receive the initial response, not the full lifetime of the request.

PR #962, opened 2018-12-21
switch travis build to xenial

brharrington:
Also increase memory for build. There have been a few
sporadic build failures that may be resource related.
Xenial has fewer other services running by default.

https://blog.travis-ci.com/2018-11-08-xenial-release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
 script:
 - ./scripts/buildViaTravis.sh
 sudo: required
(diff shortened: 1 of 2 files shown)


brharrington:
Looks like oraclejdk8 is not available on xenial. Will need to wait until we can move to openjdk11 for the build.

brharrington:
Will use osx for now to get more consistent font rendering between the most common developer machines and CI. See #973.

PR #967, opened 2018-12-21
switch to RobotoMono font

brharrington:
The Lucida fonts are not included with OpenJDK and have been
removed from OracleJDK in version 11. The RobotoMono family
is Apache licensed and will now be used as the default to get
a more consistent experience across JDK versions.


brharrington:
For more context: https://www.oracle.com/technetwork/java/javase/11-relnote-issues-5012449.html#JDK-8191522

PR #983, opened 2019-01-24
Throttle calls to CloudWatch

zimmermatt:
We're hitting CloudWatch rate limits on a regular basis. However, the
AWS limits should be sufficient for our overall per second call rate in
the majority of cases. The current pattern of calls has bursts when the
`Tick` message kicks off a collection, which causes the call rate to
spike above the per second limit. This commit introduces call rate
limiting to smooth out the request pattern.

The Akka documentation for throttling request/response actor
communication is incomplete. Through iteration playing around in a local
toy app, I arrived at the implementation herein and confirmed that it
works as expected.

### DIFF
--- a/atlas-poller-cloudwatch/src/main/resources/reference.conf
+++ b/atlas-poller-cloudwatch/src/main/resources/reference.conf
@@ -46,6 +46,12 @@ atlas {
     // How long to cache datapoints to avoid gaps
     cache-ttl = 20 minutes
 
+    metrics-get-buffer-size = 1000
+    metrics-get-max-rate-per-second = 400
+
+    metrics-list-buffer-size = 1000
+    metrics-list-max-rate-per-second = 25
+
     // Class to use for mapping AWS dimensions to a tag map for use in Atlas
     tagger = {
       class = "com.netflix.atlas.cloudwatch.NetflixTagger"
--- a/atlas-poller-cloudwatch/src/main/resources/reference.conf
+++ b/atlas-poller-cloudwatch/src/main/resources/reference.conf
@@ -49,9 +49,6 @@ atlas {
     metrics-get-buffer-size = 1000
     metrics-get-max-rate-per-second = 400
 
-    metrics-list-buffer-size = 1000
-    metrics-list-max-rate-per-second = 25
-
     // Class to use for mapping AWS dimensions to a tag map for use in Atlas
     tagger = {
       class = "com.netflix.atlas.cloudwatch.NetflixTagger"
--- a/atlas-poller-cloudwatch/src/main/resources/reference.conf
+++ b/atlas-poller-cloudwatch/src/main/resources/reference.conf
@@ -46,7 +46,7 @@ atlas {
     // How long to cache datapoints to avoid gaps
     cache-ttl = 20 minutes
 
-    metrics-get-buffer-size = 1000
+    metrics-get-buffer-size = 10
     metrics-get-max-rate-per-second = 400
 
     // Class to use for mapping AWS dimensions to a tag map for use in Atlas
--- a/atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
+++ b/atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
@@ -92,7 +92,7 @@ class CloudWatchPoller(config: Config, registry: Registry, client: AmazonCloudWa
   private val throttledMetricsGetRef = Source
     .actorRef[List[MetricMetadata]](
       config.getInt("atlas.cloudwatch.metrics-get-buffer-size"),
-      OverflowStrategy.dropNew
+      OverflowStrategy.dropHead
     )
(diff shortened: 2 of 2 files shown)


atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
```diff
@@ -81,11 +88,31 @@ class CloudWatchPoller(config: Config, registry: Registry, client: AmazonCloudWa
...
+    .toMat(Sink.foreach(message => metricsGetRef.tell(message, self)))(Keep.left)
+    .run()
+
   // Child actor for listing metrics. This will do the call using the
   // AWS SDK which is blocking and should be run in an isolated dispatcher.
   private val metricsListRef =
     context.actorOf(FromConfig.props(Props(new ListMetricsActor(client, tagger))), "metrics-list")
 
+  // Throttler to control the rate of list metrics calls in order to stay within AWS SDK limits.
+  private val throttledMetricsListRef = Source
```
brharrington: I don't think this is needed. It is limiting the rate for message to `ListMetricsActor`, but we only allow one pending list call at a time.

atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
```diff
@@ -167,15 +194,15 @@ class CloudWatchPoller(config: Config, registry: Registry, client: AmazonCloudWa
       pendingGets.addAndGet(num)
       logger.info(s"requesting data for $num metrics")
       ms.foreach { m =>
-        metricsGetRef ! m
+        throttledMetricsGetRef ! m
```
brharrington: If something gets dropped because it hits `metrics-get-buffer-size`, then won't `pending` increase and never decrease? This could put us in a state where we stop polling because the `pending` size is too high.

If `ms.size` is much larger than `metrics-get-buffer-size`, then it seems likely we would dropping a lot of the messages before they are processed. Since this loop will be really fast compared to the throttled output, most of the time I would expect the first `metrics-get-buffer-size` or so to get processed and the rest would be dropped each time.

zimmermatt: Good point. That is possible. I'll update with the solution that has stronger guarantees for the list of available metrics.

On a side note... I found as I was playing around with the solution that, in practice, it's difficult to trigger this scenario, however. I think the combination of actors having an unbounded mailbox and other buffering is at play in that behavior. I had to go so far as to configure a `NonBlockingBoundedMailbox` with a capacity of `1`. Even then, it goes for quite some time without dropping any messages.

zimmermatt: Interesting... My toy app mimics the interactions that `CloudWatchPoller` and `GetMetricsActor` have. That is, messages sent to one result in messages sent to the other with a `tell` call in the `Sink.foreach`. That toy app behaves the way I describe.

atlas-poller-cloudwatch/src/main/resources/reference.conf
```diff
@@ -46,6 +46,9 @@ atlas {
     // How long to cache datapoints to avoid gaps
     cache-ttl = 20 minutes
 
+    metrics-get-buffer-size = 1000
```
brharrington: Given the change to accept the list, this buffer doesn't need to be very large.

atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
```diff
@@ -81,6 +88,17 @@ class CloudWatchPoller(config: Config, registry: Registry, client: AmazonCloudWa
   private val metricsGetRef =
     context.actorOf(FromConfig.props(Props(new GetMetricActor(client))), "metrics-get")
 
+  // Throttler to control the rate of get metrics calls in order to stay within AWS SDK limits.
+  private val throttledMetricsGetRef = Source
+    .actorRef[List[MetricMetadata]](
+      config.getInt("atlas.cloudwatch.metrics-get-buffer-size"),
+      OverflowStrategy.dropNew
```
brharrington: Since it contains the complete list, it seems like it would be better to keep the newest complete list rather than an old one in the queue.

atlas-poller-cloudwatch/src/main/resources/reference.conf
zimmermatt: agreed

atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
zimmermatt: good point

PR #987, opened 2019-01-29
add rolling-mean operator

brharrington:
This is meant as an alternative to `:trend` that fixes a
number of issues with that operator. Specifically:

1. The denominator for the average is the number of actual
   values, that is non-NaN entries, within the rolling buffer.
   The `:trend` operator always uses the window size which
   can create confusing drops because `NaN` values are
   effectively 0.
2. The minimum number of values permitted before emitting
   a mean can be specified by the user.
3. It is more consistent with other stateful operators in
   that it works on a window size relative to the step
   interval rather than a fixed time duration.
4. It is more consistent with similar operators in other
   tools such as the `rolling_mean` function provided by
   Pandas.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineRollingMean.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineRollingMean.scala
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2014-2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineRollingMean.scala
```diff
@@ -0,0 +1,80 @@
...
+      sum -= removed
+      count -= 1
+    }
+
+    if (!JDouble.isNaN(v)) {
+      sum += v
+      count += 1
+    }
+
+    if (count >= minNumValues) sum / count else Double.NaN
```
dmuino: Can minNumValues be 0? We could potentially divide-by zero in that case.

brharrington: I added a check

PR #1001, opened 2019-03-01
Improve CloudWatch metric lag handling

zimmermatt:
This is the first of potentially multiple commits to improve handling
of CloudWatch metric lag. This first commit makes the time range of
the query configurable and adds a metric to track the age (in
CloudWatch periods) of the latest datapoint returned. This will enable
assessing the distribution of ages across the namespaces and metrics
collected. Those data will influence the approach for datapoints that
are older than Atlas will accept.

### DIFF
--- a/atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
+++ b/atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
@@ -21,6 +21,7 @@ import java.util.Date
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicLong
 import java.util.concurrent.atomic.AtomicReference
+import java.util.function.LongFunction
 
 import akka.actor.Actor
 import akka.actor.ActorRef
@@ -37,7 +38,9 @@ import com.amazonaws.services.cloudwatch.model.StandardUnit
 import com.github.benmanes.caffeine.cache.Caffeine
 import com.netflix.atlas.poller.Messages
 import com.netflix.spectator.api.Functions
+import com.netflix.spectator.api.Id
 import com.netflix.spectator.api.Registry
+import com.netflix.spectator.api.histogram.BucketCounter
 import com.netflix.spectator.api.patterns.PolledMeter
 import com.typesafe.config.Config
 import com.typesafe.scalalogging.StrictLogging
@@ -86,7 +89,17 @@ class CloudWatchPoller(config: Config, registry: Registry, client: AmazonCloudWa
   // Child actor for getting the data for a metric. This will do the call using the
   // AWS SDK which is blocking and should be run in an isolated dispatcher.
   private val metricsGetRef =
-    context.actorOf(FromConfig.props(Props(new GetMetricActor(client))), "metrics-get")
+    context.actorOf(
+      FromConfig.props(
+        Props(
+          classOf[GetMetricActor],
+          client,
+          registry,
+          buildBucketCounterCache(registry, categories)
+        )
+      ),
+      "metrics-get"
+    )
 
   // Throttler to control the rate of get metrics calls in order to stay within AWS SDK limits.
   private val throttledMetricsGetRef = Source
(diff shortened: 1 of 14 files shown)


atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/GetMetricActor.scala
```diff
@@ -53,11 +68,31 @@ class GetMetricActor(client: AmazonCloudWatch) extends Actor with StrictLogging
...
     }
   }
+
+  /**
+    * Record how many periods back from now the latest returned datapoint is. Though not perfect,
+    * this will give a reasonable approximation of data latency across the collected metrics.
+    */
+  private def recordLag(now: Instant, maybeTimestamp: Option[Date], m: MetricMetadata): Unit = {
+    val mostRecentDatapointTimestamp = maybeTimestamp.map(_.toInstant).getOrElse(Instant.EPOCH)
+    val lagDuration = Duration.between(mostRecentDatapointTimestamp, now)
```
brharrington: Why does this compare to now? Since there is an offset applied to the end time for the data being fetched it isn't really recording the lag in cloudwatch unless it happens to be slower than our configured offset. A better option might be to always fetch the data with an end time of now and use the latest value received for measuring the lag, but use the latest according to the offset as the datapoint value.

atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/GetMetricActor.scala
```diff
@@ -53,11 +68,31 @@ class GetMetricActor(client: AmazonCloudWatch) extends Actor with StrictLogging
...
+
+    val lagSeconds = lagDuration.toMillis / 1000L
+    val periodLag = lagSeconds / m.category.period
+    val id = basePeriodLagId
+      .withTag("cwMetricName", m.definition.name)
+      .withTag("cwNamespace", m.category.namespace)
+      .withTag("periodSeconds", m.category.period.toString)
```
brharrington: Since this the period is fixed in the config, is it really needed on the metric?

atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
```diff
@@ -288,6 +301,41 @@ object CloudWatchPoller {
     cls.getConstructor(classOf[Config]).newInstance(cfg).asInstanceOf[Tagger]
   }
 
+  val PeriodLagIdName: String = "atlas.cloudwatch.periodLag"
+
+  private def buildBucketCounterCache(
```
brharrington: It seems like a basic distribution summary of the period count would be a lot more straightforward.

atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/GetMetricActor.scala
zimmermatt: Good point. This is a bit of bias sneaking in. All of the metrics I've ad-hoc analyzed thus far have been _at least _ 2 periods late, so I didn't bother with that additional complexity. Fundamentally, however, you're right. To get the best accuracy, I should either compare against the offset or get all data for the lag calculation.

I'll post an update in a bit.

atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/GetMetricActor.scala
zimmermatt: I anticipate needing to use graphs of this to communicate the issues I'm seeing to a variety of stakeholders. The overall dimensionality is relatively low, so I'm favoring ease and understanding over redundancy. Having that dimension directly associated will ease creating those graphs and enable others to explore the data with better understanding even without a lot of pre-existing context. We can always remove it later, once no longer useful.

atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
zimmermatt: Given the communication needs, I want to have the specificity this offers.

Issue #1002, opened 2019-03-06
Import additional RDS/Aurora cloudwatch metrics


asher:
Please add the following Aurora specific cloudtrail metrics to the list of RDS metrics imported to atlas:

`AuroraReplicaLag`
`AuroraReplicaLagMaximum`
`FreeLocalStorage`

zimmermatt:
`FreeLocalStorage` is already available as `aws.rds.diskFree`.

https://github.com/Netflix/atlas/blob/master/atlas-poller-cloudwatch/src/main/resources/rds.conf#L32-L34

PR #1003, opened 2019-03-07
Collect additional aurora replica metrics

zimmermatt:
Add collection of `AuroraReplicaLag` and `AuroraReplicaLagMaximum`
(issue #1002).

### DIFF
--- a/atlas-poller-cloudwatch/src/main/resources/rds.conf
+++ b/atlas-poller-cloudwatch/src/main/resources/rds.conf
@@ -64,6 +64,34 @@ atlas {
           name = "ReplicaLag"
           alias = "aws.rds.replicaLag"
           conversion = "max"
+          tags = [
+            {
+              key = "engine"
+              value = "other"
+            }
+          ]
+        },
+        {
+          name = "AuroraReplicaLag"
+          alias = "aws.rds.replicaLag"
+          conversion = "max"
+          tags = [
+            {
+              key = "engine"
+              value = "aurora"
+            }
+          ]
+        },
+        {
+          name = "AuroraReplicaLagMaximum"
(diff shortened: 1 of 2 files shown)


atlas-poller-cloudwatch/src/main/resources/rds.conf
```diff
@@ -64,6 +64,34 @@ atlas {
...
+            {
+              key = "engine"
+              value = "other"
+            }
+          ]
+        },
+        {
+          name = "AuroraReplicaLag"
+          alias = "aws.rds.replicaLag"
+          conversion = "max"
```
brharrington: According to the docs these are in milliseconds, they should get normalized to a base unit of seconds. Not sure why they are so inconsistent with units, `ReplicaLag` and `AuroraBinlogReplicaLag` are in seconds.

zimmermatt: The `Unit` in the datapoints come back as `Milliseconds`, so the normalization is already handled.

Added a test to ensure this.

Example responses:
```
{
    "Label": "AuroraReplicaLag",
    "Datapoints": [
        {
            "Timestamp": "2019-03-07T18:32:00Z",
            "SampleCount": 1.0,
            "Sum": 15.867,
            "Minimum": 15.867,
            "Maximum": 15.867,
            "Unit": "Milliseconds"
        }
    ]
}
```
```
{
    "Label": "AuroraReplicaLagMaximum",
    "Datapoints": [
        {
            "Timestamp": "2019-03-07T18:33:00Z",
            "SampleCount": 1.0,
            "Sum": 20.312999725341797,
            "Minimum": 20.312999725341797,
            "Maximum": 20.312999725341797,
            "Unit": "Milliseconds"
        }
    ]
}
```

Issue #1005, opened 2019-03-15
Missing aws lambda metrics


JonathanBond:
Could you please add the aws lambda metrics IteratorAge, DeadLetterErrors, ConcurrentExecution and UnreservedConcurrentExecutions.  Thanks, JB

zimmermatt:
Discussed with @JonathanBond. He's going to set up a sample, so we can verify the dimensions on these.

bondj:
DeadLetterErrors is {
    "view": "timeSeries",
    "stacked": false,
    "metrics": [
        [ "AWS/Lambda", "DeadLetterErrors", "FunctionName", "multiRegion-function1-prod-prod", "Resource", "multiRegion-function1-prod-prod:$LATEST" ]
    ],
    "region": "us-east-1"
}

PR #1008, opened 2019-03-27
Collect additional lambda metrics

zimmermatt:
Add collection of
* `ConcurrentExecutions`
* `UnreservedConcurrentExecutions`
* `DeadLetterErrors`
* `IteratorAge`

Issue #1005

### DIFF
--- a/atlas-poller-cloudwatch/src/main/resources/lambda.conf
+++ b/atlas-poller-cloudwatch/src/main/resources/lambda.conf
@@ -7,6 +7,36 @@ atlas {
       namespace = "AWS/Lambda"
       period = 1m
 
+      metrics = [
+        {
+          name = "ConcurrentExecutions"
+          alias = "aws.lambda.concurrentExecutions"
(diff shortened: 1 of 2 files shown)


atlas-poller-cloudwatch/src/main/resources/lambda.conf
```diff
@@ -7,32 +7,72 @@ atlas {
       namespace = "AWS/Lambda"
       period = 1m
 
+      metrics = [
+        {
+          name = "ConcurrentExecutions"
+          alias = "aws.lambda.concurrentExecutions"
+          conversion = "sum,rate"
```
brharrington: This doesn't seem like it should be a rate. A gauge seems more appropriate.

zimmermatt: Yeah. Re-reading the docs, I agree. I read too much into "count" as the unit.

PR #1010, opened 2019-03-28
Fix lambda config

zimmermatt:
The code that loads `MetricCategory` instances from the config expects
the `dimension` field, even if it's empty. This was causing the service
to silently fail.

In addition to fixing the `lambda` config, I've added tests that load
the production config to improve the chances we'll catch this at build
time. I've also added logging, so the problem will be evident in the
logs.

### DIFF
--- a/atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
+++ b/atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
@@ -266,7 +266,7 @@ class CloudWatchPoller(config: Config, registry: Registry, client: AmazonCloudWa
   }
 }
 
-object CloudWatchPoller {
+object CloudWatchPoller extends StrictLogging {
 
   case object Flush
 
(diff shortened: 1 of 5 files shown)


atlas-poller-cloudwatch/src/main/scala/com/netflix/atlas/cloudwatch/CloudWatchPoller.scala
```diff
@@ -286,19 +286,31 @@ object CloudWatchPoller {
...
-      val cfg = config.getConfig(s"atlas.cloudwatch.$name")
-      MetricCategory.fromConfig(cfg)
+    val categories = try {
+      config.getStringList("atlas.cloudwatch.categories").asScala.map { name =>
+        val cfg = config.getConfig(s"atlas.cloudwatch.$name")
+        MetricCategory.fromConfig(cfg)
+      }
+    } catch {
+      case t: Throwable =>
+        logger.error("Problem loading categories", t)
```
brharrington: I think a better solution here is to to it in the supervision strategy for the manager so that anything that causes the initialization to fail would get logged. See #1011.

zimmermatt: agreed.

PR #1012, opened 2019-03-29
Ensure git tags are available for build

zimmermatt:
`git fetch --unshallow` was being used to get tags to set the artifact
version. This has started failing on one of the builds. This commit
switches to explicitly fetching tags via `git fetch --tags`.


brharrington:
Running the steps travis is doing locally, I can reproduce the behavior. This seems to work for me:

```
git fetch --unshallow --tags
```

PR #1018, opened 2019-04-04
support generated dataset for eval

### LINKED ISSUE #798: Support raw datapoint input for atlas-eval
There is a growing use-case around running atlas stack language evaluation on timeseries data that does not reside in Atlas and may not even be reported from the on instance client. It would really benefit us to use the atlas-eval library in a mode where the input form is a Stream of DataPoint messages, where DataPoint could look like `(ts: Long, value: Double, tags: Map[String, String])` 

This will help use cases targeted at alerting on streaming data, where users want to run some form of alerting but can't today, unless they store this data into Atlas. From an Atlas-Alerting perspective, this will help us on-board other time series data sources, while still adhering to the stack language as the DSL standard.

brharrington:
Adds a `evaluator.createDatapointProcessor(DataSources)` method
that can be used to process a generated set of datapoints rather
than getting data from an LWC cluster. This can be used for tests
or for using Atlas expressions over arbitrary event streams.

Fixes #798.

/cc @vshirikian @ruchirj

### DIFF
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
@@ -45,6 +45,10 @@ import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -103,6 +107,28 @@ public final class Evaluator extends EvaluatorImpl {
    */
   public Processor<DataSources, MessageEnvelope> createStreamsProcessor() {
     return createStreamsProcessorImpl();
+  }
+
+  /**
+   * Creates a processor that evaluates a fixed set of expressions against the stream of
+   * time grouped datapoints. The caller must ensure:
+   *
+   * <ul>
+   *   <li>All data sources have the same step size. This is required because there is a
+   *       single input datapoint stream with a fixed step.</li>
+   *   <li>There should be a single group for each step interval with timestamps aligned to
+   *       the step boundary.</li>
+   *   <li>Operations are feasible for the input. Time based operators may or may not work
+   *       correctly depending on how the timestamps for the input groups are determined.
+   *       For use-cases where the the timestamps are generated arbitrarily, such as incrementing
+   *       for each group once a certain number of datapoints is available, then operators
+   *       based on a clock such as {@code :time} will have bogus values based on the arbitrary
+   *       times for the group. If the data is a snapshot of an actual stream, then time based
+   *       operators should work correctly.</li>
+   * </ul>
+   */
+  public Processor<DatapointGroup, MessageEnvelope> createDatapointProcessor(DataSources sources) {
+    return createDatapointProcessorImpl(sources);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


brharrington:
Example of usage from java using rxjava2:

```java
import akka.actor.ActorSystem;
import akka.stream.ActorMaterializer;
import com.netflix.spectator.api.NoopRegistry;
import com.typesafe.config.Config;
import com.typesafe.config.ConfigFactory;
import io.reactivex.Flowable;
import org.reactivestreams.Processor;

import java.time.Duration;
import java.util.Iterator;
import java.util.LinkedHashMap;
import java.util.Map;
import java.util.Random;

/**
 * Simulates performing an evaluation over a stream of events using arbitrary batches of
 * 5k events and assigning a synthetic timestamp that is effectively the batch number.
 */
public class Example {

  public static void main(String[] args) {

    // Setup evaluator
    Config config = ConfigFactory.load();
    ActorSystem system = ActorSystem.create();
    ActorMaterializer mat = ActorMaterializer.create(system);
    Evaluator evaluator = new Evaluator(config, new NoopRegistry(), system);

    // The step for this example is 1 as we are basically just using the timestamps as a
    // batch number
    Duration step = Duration.ofSeconds(1L);

    // Set of datasources is fixed, would be generated from something like an ACA or
    // alert config
    Evaluator.DataSources sources = Evaluator.DataSources.of(
        new Evaluator.DataSource("0", step, uri("name,cpu,:eq,:avg,$name,:legend")),
        new Evaluator.DataSource("1", step, uri("name,requests,:eq,:sum,$name,:legend")),
        new Evaluator.DataSource("2", step, uri("name,latency,:eq,:max,$name,:legend")),
        new Evaluator.DataSource("3", step, uri("name,errors,:eq,:sum,(,node,),:by,$node,:legend"))
    );

    // Process data
    Processor<Evaluator.DatapointGroup, Evaluator.MessageEnvelope> processor =
        evaluator.createDatapointProcessor(sources);
    Flowable.fromIterable(generateData())
        .buffer(5000)
        .zipWith(
            Flowable.range(0, Integer.MAX_VALUE),
            (ds, i) -> new Evaluator.DatapointGroup(i * 1000, ds))
        .subscribe(processor);
    Flowable.fromPublisher(processor)
        .take(100)
        .blockingIterable()
        .forEach(msg -> System.out.println(msg.getId() + " -> " + msg.getMessage().toJson()));

    // Shutdown actor system
    system.terminate();
  }

  private static String uri(String expr) {
    return "http://localhost/api/v1/graph?q=" + expr;
  }

  //////////////////////////////////////////////////////////////////
  // Generate some random data for the example...

  private static Map<String, String> map(String... vs) {
    Map<String, String> tags = new LinkedHashMap<>();
    for (int i = 0; i < vs.length; i += 2) {
      tags.put(vs[i], vs[i + 1]);
    }
    return tags;
  }

  private static Iterable<Evaluator.Datapoint> generateData() {
    return () -> new Iterator<>() {
      private int i = 0;
      private Random random = new Random();

      @Override public boolean hasNext() {
        return true;
      }

      @Override
      public Evaluator.Datapoint next() {
        switch (i++ % 7) {
          case 0: // cpu
            return new Evaluator.Datapoint(
                map("name", "cpu", "node", "" + (i / 4 % 8)),
                random.nextDouble() * 100.0);
          case 1: // latency
            return new Evaluator.Datapoint(
                map("name", "latency", "node", "" + (i / 4 % 8)),
                random.nextDouble());
          case 2: // RPS, assume each request is a separate datapoint. Weight it so there are
          case 3: // more requests than errors
          case 4:
          case 5:
            return new Evaluator.Datapoint(
                map("name", "requests", "node", "" + (i / 4 % 8)),
                1.0);
          default: // errors
            return new Evaluator.Datapoint(
                map("name", "errors", "node", "" + (i / 4 % 8)),
                1.0);
        }
      }
    };
  }
}
```

ruchirj:
👍 looks good, thanks!

copperlight:
The example requires a snapshot build until a new release can be cut that includes the changes:

```
repositories {
    maven {
        url 'http://oss.jfrog.org/oss-snapshot-local'
    }
}

dependencies {
    compile 'io.reactivex.rxjava2:rxjava:2.+'
    compile 'com.netflix.atlas_v1:atlas-eval_2.12:1.6.0-SNAPSHOT'
}
```

PR #1025, opened 2019-04-16
Add rolling-sum stateful operator

dmuino:
Sum of the values within a specified window. The sum will only be emitted
if there are at least a minimum number of actual values (not `NaN`) within
the window. Otherwise `NaN` will be emitted for that time period.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineRollingSum.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineRollingSum.scala
@@ -18,8 +18,7 @@ package com.netflix.atlas.core.algorithm
 import com.netflix.atlas.core.util.Math
 
 /**
-  * Sum of the values within a moving window of the input. The denominator is the number
-  * of values (non-NaN entries) in the rolling buffer.
+  * Sum of the values within a moving window of the input.
   *
   * @param buf
   *     Rolling buffer to keep track of the input for a given window.
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineRollingSum.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineRollingSum.scala
@@ -22,16 +22,10 @@ import com.netflix.atlas.core.util.Math
   *
   * @param buf
   *     Rolling buffer to keep track of the input for a given window.
-  * @param minNumValues
-  *     Minimum number of values that must be present within the buffer for a sum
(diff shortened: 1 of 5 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineRollingSum.scala
```diff
@@ -0,0 +1,79 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.atlas.core.algorithm
+
+import com.netflix.atlas.core.util.Math
+
+/**
+  * Sum of the values within a moving window of the input. The denominator is the number
```
brharrington: denominator? I'm guessing this was copied from `rolling-mean`, but doesn't apply for a sum.

atlas-core/src/main/scala/com/netflix/atlas/core/model/StatefulVocabulary.scala
```diff
@@ -308,6 +308,55 @@ object StatefulVocabulary extends Vocabulary {
     override def examples: List[String] = List("name,sps,:eq,:sum,5,3")
   }
 
+  case object RollingSum extends SimpleWord {
```
brharrington: This would need to get added to the list of words for the vocabulary (see [L33](https://github.com/Netflix/atlas/blob/master/atlas-core/src/main/scala/com/netflix/atlas/core/model/StatefulVocabulary.scala#L33)).

dmuino: Thanks, fixed.

atlas-core/src/main/scala/com/netflix/atlas/core/algorithm/OnlineRollingSum.scala
```diff
@@ -0,0 +1,78 @@
...
+package com.netflix.atlas.core.algorithm
+
+import com.netflix.atlas.core.util.Math
+
+/**
+  * Sum of the values within a moving window of the input.
+  *
+  * @param buf
+  *     Rolling buffer to keep track of the input for a given window.
+  * @param minNumValues
```
brharrington: Thinking about this some more, is this parameter needed? In the case of `rolling-mean` it is used to avoid what looks to the user as drops due to missing data. For the sum I wonder if it would be better to just always show the full sum.

PR #1038, opened 2019-05-02
update base version to 1.7.x

brharrington:


brharrington:
Confirmed 1.7.0-SNAPSHOT was produced.

PR #1060, opened 2019-06-16
match cloudwatch usage of `canceled`

copperlight:


matschaffer:
Lgtm 😄

atlas-poller-cloudwatch/src/main/resources/swf.conf
```diff
@@ -53,7 +53,7 @@ atlas {
           tags = [
             {
               key = "status"
-              value = "cancelled"
+              value = "canceled"
```
brharrington: For consistency, `ActivityTasksCanceled` should also be updated.

PR #1092, opened 2019-10-02
update travis to use Mac OS 10.14

brharrington:
Looks like the font rendering in Mac OS changed so we get
diffs when running locally. See if this causes travis to
fail so we can update to get consistent rendering again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: scala
 matrix:
   include:
-  - jdk: openjdk11
+  - jdk: openjdk12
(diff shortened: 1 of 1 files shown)


brharrington:
The font rendering issue actually seems to be some difference between OpenJDK 11.0.2 (`11.0.2-open`) and Zulu 11.0.4 (`11.0.4-zulu`).

brharrington:
Most likely the change was caused by: https://bugs.openjdk.java.net/browse/JDK-8218854

Issue #1007, opened 2019-03-27
Improve tick label selection for logarithmic axis


brharrington:
```
/api/v1/graph?q=1,1000,1e6&scale=log
```

![logscale](https://user-images.githubusercontent.com/1289028/55112440-d8430500-5099-11e9-8c7b-d3b5c83c4f09.png)

brharrington:
```
/api/v1/graph?q=1e4,1e5,1e6,1.766e6&scale=log
```

Spacing is a bit odd for this example

PR #1093, opened 2019-10-08
even out tick distance in graph if value range is big

jfz:
#1007


atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/Ticks.scala
```diff
@@ -235,6 +243,45 @@ object Ticks {
...
+      //val v = math.pow(10, curr)
+      val label = UnitPrefix.format(v, "%.0f%s")
+      ticks += ValueTick(v, 0.0, (curr - min) % logMajorStepSize == 0, Some(label))
+      curr += 1
+    }
+
+    ticks.result()
+  }
+
+  def getLogMajorStepSize(min: Double, max: Double, n: Int): Int = {
```
brharrington: Should this be private?

atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/Ticks.scala
```diff
@@ -235,6 +243,45 @@ object Ticks {
...
+    val min = logFloor(v1).toInt
+    val max = logCeil(v2).toInt
+    val logMajorStepSize = getLogMajorStepSize(min, max, n)
+
+    val ticks = List.newBuilder[ValueTick]
+    var curr = min
+    while (curr <= max) {
+      //show tick for 0 but not 1(10^0) if lower boundary is 0, because they are too close in log scale
+      val v = if (v1 == 0 && curr == 0) 0 else math.pow(10, curr)
+      //val v = math.pow(10, curr)
```
brharrington: Remove commented out code

brharrington:
This doesn't work correctly for negative values.

**Before**

old

**After**

new

PR #1098, opened 2019-10-22
support lwcapi instance lookup via edda 

jfz:
Do lwcapi instance lookup via Edda, if the uri looks like edda (contains "/autoScalingGroups")

### DIFF
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaSource.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaSource.scala
@@ -26,6 +26,7 @@ import akka.stream.scaladsl.Compression
 import akka.stream.scaladsl.Flow
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
+import com.fasterxml.jackson.annotation.JsonProperty
 import com.netflix.atlas.json.Json
 import com.typesafe.scalalogging.StrictLogging
 
@@ -48,7 +49,6 @@ private[stream] object EurekaSource extends StrictLogging {
     */
   def apply(eurekaUri: String, context: StreamContext): Source[GroupResponse, NotUsed] = {
 
-    val useVipFormat = eurekaUri.contains("/vips/")
     val headers =
       List(Accept(MediaTypes.`application/json`), `Accept-Encoding`(HttpEncodings.gzip))
     val request = HttpRequest(HttpMethods.GET, eurekaUri, headers)
@@ -58,7 +58,7 @@ private[stream] object EurekaSource extends StrictLogging {
       .via(context.httpClient("eureka"))
       .flatMapConcat {
         case Success(res: HttpResponse) if res.status == StatusCodes.OK =>
-          parseResponse(eurekaUri, res, useVipFormat)
+          parseResponse(eurekaUri, res)
         case Success(res: HttpResponse) =>
           logger.warn(s"eureka refresh failed with status ${res.status}: $eurekaUri")
           res.discardEntityBytes(context.materializer)
@@ -76,8 +76,7 @@ private[stream] object EurekaSource extends StrictLogging {
 
   private def parseResponse(
     uri: String,
-    res: HttpResponse,
-    vipFormat: Boolean
+    res: HttpResponse
   ): Source[GroupResponse, Any] = {
     unzipIfNeeded(res)
       .reduce(_ ++ _)
@@ -88,11 +87,19 @@ private[stream] object EurekaSource extends StrictLogging {
       }
       .filter(_.nonEmpty)
       .map { bs =>
-        if (vipFormat)
+        if(uri.contains("/autoScalingGroups"))
+          decodeEddaResponse(bs.toArray).copy(uri = uri)
+        else if (uri.contains("/vips/"))
           Json.decode[VipResponse](bs.toArray).copy(uri = uri)
         else
           Json.decode[AppResponse](bs.toArray).copy(uri = uri)
       }
+  }
+
+  private def decodeEddaResponse(ba: Array[Byte]) = {
+    val responses = Json.decode[List[EddaResponse]](ba)
+    require(responses != null && !responses.isEmpty, "EddaResponse list cannot be empty")
+    responses(0)
   }
 
   //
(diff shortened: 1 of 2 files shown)


atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaSource.scala
```diff
@@ -88,13 +87,21 @@ private[stream] object EurekaSource extends StrictLogging {
...
+        if(uri.contains("/autoScalingGroups"))
+          decodeEddaResponse(bs.toArray).copy(uri = uri)
+        else if (uri.contains("/vips/"))
           Json.decode[VipResponse](bs.toArray).copy(uri = uri)
         else
           Json.decode[AppResponse](bs.toArray).copy(uri = uri)
       }
   }
 
+  private def decodeEddaResponse(ba: Array[Byte]) = {
```
brharrington: Avoid procedure syntax.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaSource.scala
```diff
@@ -88,13 +87,21 @@ private[stream] object EurekaSource extends StrictLogging {
...
+    val responses = Json.decode[List[EddaResponse]](ba)
+    require(responses != null && !responses.isEmpty, "EddaResponse list cannot be empty")
+    responses(0)
```
brharrington: It should merge the results from all clusters instead of picking the first. If there was a deployment going on it would ignore either the old or new instances right now.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaSource.scala
```diff
@@ -88,13 +87,21 @@ private[stream] object EurekaSource extends StrictLogging {
...
+        else if (uri.contains("/vips/"))
           Json.decode[VipResponse](bs.toArray).copy(uri = uri)
         else
           Json.decode[AppResponse](bs.toArray).copy(uri = uri)
       }
   }
 
+  private def decodeEddaResponse(ba: Array[Byte]) = {
+    val responses = Json.decode[List[EddaResponse]](ba)
+    require(responses != null && !responses.isEmpty, "EddaResponse list cannot be empty")
```
brharrington: It shouldn't be `null`, but I'm not sure it should throw for an empty list. That is most likely a miss-configuration if the clusters were deleted. For now it should probably follow the Eureka behavior to just be an empty group.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaSource.scala
```diff
@@ -123,6 +130,24 @@ private[stream] object EurekaSource extends StrictLogging {
   case class App(name: String, instance: List[Instance]) {
     require(instance != null, "instance cannot be null")
   }
+  //the json field name "instances" conflicts with method name, need to explicitly map it with annotation
+  case class EddaResponse(uri: String, @JsonProperty("instances") eddaInstances: List[EddaInstance]) extends GroupResponse {
+    require(eddaInstances != null && !eddaInstances.isEmpty, "eddaInstances cannot be empty")
```
brharrington: The set of instances for a cluster could be captured as an empty set during a deployment.

atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/EurekaSourceSuite.scala
```diff
@@ -177,4 +177,29 @@ class EurekaSourceSuite extends FunSuite {
...
     }
   }
+
+  test("handles edda uri") {
+    val uri = "http://edda/api/v2/group/autoScalingGroups;cluster=atlas_lwcapi-main;_expand"
+    val res = run(uri, Success(mkResponse(eddaResponseStr)))
+    assert(res.uri === uri)
+    assert(res.instances.size === 2)
+    assert(res.instances.map(_.instanceId).toSet === Set("id1", "id2"))
+    assert("http://1.2.3.4:7101" === res.instances(0).substitute("http://{local-ipv4}:{port}"))
```
brharrington: It would be more idiomatic to use `instances.head`

atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/EurekaSourceSuite.scala
```diff
@@ -177,4 +177,29 @@ class EurekaSourceSuite extends FunSuite {
...
+  }
+
+  val eddaResponseStr =
+    """[
```
brharrington: Can you add a test case that has multiple server groups in the response?

brharrington:
Looks like build failed due to formatting, can you run `make format` to clean it up?

atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/EurekaSourceSuite.scala
jfz: other elements need to be validated too, so I think it would be more consistent this way

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaSource.scala
```diff
@@ -131,17 +131,19 @@ private[stream] object EurekaSource extends StrictLogging {
     require(instance != null, "instance cannot be null")
   }
   //the json field name "instances" conflicts with method name, need to explicitly map it with annotation
-  case class EddaResponse(uri: String, @JsonProperty("instances") eddaInstances: List[EddaInstance]) extends GroupResponse {
-    require(eddaInstances != null && !eddaInstances.isEmpty, "eddaInstances cannot be empty")
```
brharrington: I think you can probably keep the `null` check.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaSource.scala
```diff
@@ -96,10 +96,10 @@ private[stream] object EurekaSource extends StrictLogging {
       }
   }
 
-  private def decodeEddaResponse(ba: Array[Byte]) = {
+  private def decodeEddaResponse(ba: Array[Byte]): EddaResponse = {
     val responses = Json.decode[List[EddaResponse]](ba)
-    require(responses != null && !responses.isEmpty, "EddaResponse list cannot be empty")
-    responses(0)
+    require(responses != null, "EddaResponse list cannot be null")
+    EddaResponse(null, responses.filter(_.eddaInstances != null).map(_.eddaInstances).flatten)
```
brharrington: The map then flatten cold be simplified to just flatMap, something like:

```responses.flatMap(_.eddaInstances)```

File: atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/EurekaSourceSuite.scala [hunk 9]
```diff
@@ -200,6 +238,31 @@ class EurekaSourceSuite extends FunSuite {
...
+      |        "privateIpAddress": "1.2.3.4",
+      |        "instanceId": "id1"
+      |      },
+      |      {
+      |        "privateIpAddress": "1.2.3.5",
+      |        "instanceId": "id2"
+      |      }
+      |    ]
+      |  },
+      |  {}
```
brharrington: The edda payload should include the `instances` key with an empty list.

jfz: let me fix this accordingly

PR #1118, opened 2019-12-19
allow updating lwc stream subscriptions via web socket API

jfz:
allow updating lwc stream subscriptions via web socket API


atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/WebSocketSessionManager.scala
```diff
@@ -0,0 +1,106 @@
...
+            queueHandler.offer(
+              DiagnosticMessage.error(s"cannot decode json: ${exprStr}, error: ${e.getMessage}")
+            )
+          }
+          //need to pull here: no push out so no pull from downstream
+          pull(in)
+          return // stop here on json parsing error
+        }
+
+        val errors = subscribeFunc(streamId, newExpressions) //update subscription here
```
brharrington: Could we move all this stuff into the try block and then do the push/pull in one place as part of a finally block? If needed the JSON check could still be isolated by have another block looking for `JsonProcessingException` in the catch section.

PR #1123, opened 2020-01-08
Stream via lwc-api websocket 

jfz:
- Stream via lwc-api websocket: 
     1. allow subscription updates  
     2. simplify flow by avoid 2 async step (register/subscribe)
     3. data format change: no prefix for web socket api input/output json


atlas-eval/src/main/scala/com/netflix/atlas/eval/model/LwcExpression.scala
```diff
@@ -24,9 +24,9 @@ import com.netflix.atlas.json.JsonSupport
   *
   * @param expression
   *     Expression to subscribe to from LWCAPI.
-  * @param step
+  * @param frequency
   *     The step size used for this stream of data.
   */
-case class LwcExpression(expression: String, step: Long) extends JsonSupport {
+case class LwcExpression(expression: String, frequency: Long) extends JsonSupport {
```
jfz: this is to fix mismatch(step vs frequency) between client and lwc-api service side json

brharrington: Given that we use step pretty much everywhere else, ideally we would consolidate on `step`.

atlas-eval/src/main/scala/com/netflix/atlas/eval/model/LwcMessages.scala
```diff
@@ -31,7 +31,9 @@ object LwcMessages {
     */
   def parse(msg: String): AnyRef = {
     val data = Json.decode[JsonNode](msg)
-    data.get("type").asText() match {
+    val node: JsonNode = data.get("type")
+    val dataType = if (node == null) "" else node.asText()
```
brharrington: Is there a case where the type is not set? It looks like a null value would now attempt to get parsed as a DiagnosticMessage, is that intended?

Possible alternative:

```
Option(data.get("type")).map(_.asText()) match {
  case Some("...") =>
  case None => ...
}
```

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
```diff
@@ -348,51 +356,115 @@ private[stream] abstract class EvaluatorImpl(
...
+      val remoteFlow =
+        createClusterLookupFlow(context).via(createClusterStreamFlow(context))
+
+      // Streams for local
+      val localFlow = Flow[DataSources]
         .flatMapMerge(Int.MaxValue, s => Source(s.getSources.asScala.toList))
         .flatMapMerge(Int.MaxValue, s => context.localSource(Uri(s.getUri)))
-      datasources.out(1).map(_.localOnly()) ~> tmp ~> intermediateInput.in(1)
+
+      //broadcast to remote/local flow, process and merge
```
brharrington: Nit: `// Broadcast`

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
```diff
@@ -348,51 +356,115 @@ private[stream] abstract class EvaluatorImpl(
...
+      .via(context.monitorFlow("01_EurekaRefresh"))
+  }
+
+  // Streams via WebSocket API `/api/v1/subscribe`, from each instance of lwc-api cluster
+  private def createClusterStreamFlow(
+    context: StreamContext
+  ): Flow[SourcesAndGroups, ByteString, NotUsed] = {
+    Flow[SourcesAndGroups]
+      .flatMapConcat { sourcesAndGroups =>
+        //Cluster message first: need to connect before subscribe
```
brharrington: Nit: `// Cluster`

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
```diff
@@ -348,51 +356,115 @@ private[stream] abstract class EvaluatorImpl(
...
+
+  private def toExprsStr(dss: DataSources, context: StreamContext): String = {
+    val exprs = dss.getSources.asScala
+      .flatMap { dataSource =>
+        context.interpreter.eval(Uri(dataSource.getUri)).map { expr =>
+          Expression(expr.toString, dataSource.getStep.toMillis)
+        }
+      }
+      .toList
+      .sortWith(_.toString < _.toString) //sort to get a consistent string for "unique" later
```
brharrington: Wouldn't this be handled by the equals method of `DataSources`? Could we delay the conversion to a JSON string until after it has been deduped? That would also avoid the allocation of the string if it is a duplicate.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
```diff
@@ -348,51 +356,115 @@ private[stream] abstract class EvaluatorImpl(
...
+    Json.encode(exprs)
+  }
+
+  private def createWebSocketFlow(
+    instance: EurekaSource.Instance
+  ): Flow[String, String, NotUsed] = {
+    val uri = instance.substitute("ws://{local-ipv4}:{port}") + "/api/v1/subscribe"
+    val webSocketFlowOrigin = Http(system).webSocketClientFlow(WebSocketRequest(uri))
+    val webSocketFlow = Flow[String]
+      .via(StreamOps.unique) //updating subscriptions only if there's a change
```
brharrington: Nit: `// Updating`

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
```diff
@@ -348,51 +356,115 @@ private[stream] abstract class EvaluatorImpl(
...
+      .flatMapConcat {
+        case msg: TextMessage =>
+          msg.textStream.fold("")(_ + _)
+        case msg: BinaryMessage =>
+          msg.dataStream
+            .fold(ByteString.empty)(_ ++ _)
+            .map(_.decodeString(StandardCharsets.UTF_8))
+      }
+      .mapMaterializedValue(_ => NotUsed)
+    webSocketFlow
```
brharrington: Looks like you could remove this line and just not assign to a val.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
```diff
@@ -348,51 +356,115 @@ private[stream] abstract class EvaluatorImpl(
...
+      .via(StreamOps.unique) //updating subscriptions only if there's a change
+      .map(TextMessage(_))
+      .via(webSocketFlowOrigin)
+      .flatMapConcat {
+        case msg: TextMessage =>
+          msg.textStream.fold("")(_ + _)
+        case msg: BinaryMessage =>
+          msg.dataStream
+            .fold(ByteString.empty)(_ ++ _)
+            .map(_.decodeString(StandardCharsets.UTF_8))
```
brharrington: We should double check how much allocation overhead there is for converting to an intermediate string rather than just parsing direction from the ByteString. This may increase allocations quite a bit for use-cases that have a lot of data points coming through.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
jfz: Delaying conversion sounds like a better option here, will do

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
jfz: I think we should output as a String here because String is used by downstream(LwcMessages.parse) and what comes back from lwcapi is actually TextMessage. And we should avoid unnecessary conversion to ByteString and then back to String.
And one conversion is needed if lwcapi do have 2 type of Message in future.

atlas-eval/src/main/scala/com/netflix/atlas/eval/model/LwcExpression.scala
jfz: Revert this back to step and instead change SubscriptionManager.Expression to step. Is this an existing issue for the 2 step streaming?

brharrington: I think you are correct, but we should be able to mitigate it by updating the server side to treat `step` as an alias to `frequency`. See #1125.

PR #1129, opened 2020-01-16
fix delay of stopping host lookup

jfz:
There's a delay on stopping host lookup caused by EvaluationFlows.repeatWhile, use EvaluationFlows.stoppableSource instead. 
The delay can cause temporary incorrect lwc subscriptions when subscriptions change.

### DIFF
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluationFlows.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluationFlows.scala
@@ -15,8 +15,6 @@
  */
 package com.netflix.atlas.eval.stream
 
-import java.util.concurrent.atomic.AtomicBoolean
-
 import akka.Done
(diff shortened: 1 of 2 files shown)


atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluationFlows.scala
```diff
@@ -79,20 +77,6 @@ private[stream] object EvaluationFlows {
     Source.repeat(item).throttle(1, delay, 1, ThrottleMode.Shaping)
   }
 
-  /**
-    * Source that will repeat the item with the specified delay in between while the
-    * condition is still true.
-    */
-  def repeatWhile[T](item: T, delay: FiniteDuration): (Source[T, NotUsed], AtomicBoolean) = {
```
jfz: removing this since it's not used anywhere else, we can use stopableSource instead for similar use case.

PR #1137, opened 2020-01-31
emit empty DataSources for a host if no more DataSource

jfz:
Emit empty DataSources for a host if no more DataSource so that downstream can stop stream from LWC timely for this host, this is preparing for single evaluator at atlas-stream API.

### DIFF
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaGroupsLookup.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaGroupsLookup.scala
@@ -57,7 +57,7 @@ private[stream] class EurekaGroupsLookup(context: StreamContext, frequency: Fini
 
         // If there is an existing source polling Eureka, then tell it to stop. Create a
         // new instance of the flag for the next source
-        lookupTickSwitch.map(_.stop())
+        lookupTickSwitch.foreach(_.stop())
 
         val next = grab(in)
 
@@ -107,7 +107,7 @@ private[stream] class EurekaGroupsLookup(context: StreamContext, frequency: Fini
 
       override def onUpstreamFinish(): Unit = {
         completeStage()
-        lookupTickSwitch.map(_.stop())
+        lookupTickSwitch.foreach(_.stop())
       }
 
       override def onUpstreamFailure(ex: Throwable): Unit = {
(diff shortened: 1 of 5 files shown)


atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaGroupsLookup.scala
```diff
@@ -66,7 +66,8 @@ private[stream] class EurekaGroupsLookup(context: StreamContext, frequency: Fini
         if (next.getSources.isEmpty) {
           // If the Eureka based sources are empty, then just use an empty source to avoid
           // potential delays to shutting down when the upstream completes.
-          Source.empty[SourcesAndGroups]
```
jfz: Looks like the empty Source is not getting pushed previously, which should cause this part of flow stuck and previously pushed Source will not stop.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaGroupsLookup.scala
```diff
@@ -109,6 +110,11 @@ private[stream] class EurekaGroupsLookup(context: StreamContext, frequency: Fini
         lookupTickSwitch.map(_.stop())
       }
 
+      override def onUpstreamFailure(ex: Throwable): Unit = {
+        super.onUpstreamFailure(ex)
+        lookupTickSwitch.map(_.stop())
```
brharrington: `foreach` would be more appropriate here because the return type is `Unit`. It is just iterating over the items, not mapping them.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FillRemovedKeysWith.scala
```diff
@@ -0,0 +1,69 @@
...
+      private var prevKeySet = scala.collection.Set.empty[K]
+
+      override def onPush(): Unit = {
+        val map = grab(in)
+        val newMap = scala.collection.mutable.Map.empty[K, V] ++ map
+
+        val removedKeys = prevKeySet -- newMap.keySet
+        removedKeys.foreach(k => {
+          newMap.put(k, valueForRemovedKey(k))
+        })
```
brharrington: Style nitpick:

```
removedKeys.foreach { k =>
  newMap.put(k, valueForRemovedKey(k))
}
```

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EurekaGroupsLookup.scala
jfz: sounds good, thanks!

PR #1138, opened 2020-02-04
Allow updating ID of MessageEnvelope

jfz:
Allow updating ID of MessageEnvelope: need to remove streamId from message ID for single evaluator.


jfz:
> Why not just create a copy? I would like to keep all of the model object immutable if possible.
It creates extra objects, it's on the hot path which applies to every data point.

jfz:
prefer to keep it immutable over potential perf gain for now after discussion.

PR #1148, opened 2020-03-03
avoid unnecessary string conversion

brharrington:
When parsing the LWC messages converting to a string first
is unnecessary. Avoiding this results in about 30% improvement
in the throughput.


jfz:
The data from Websocket API is actually TextMessage, and got converted to ByteString, so I think, to minimize convertion, we can just fold as String from Websocket output and parse as String here? Another option is having Websocket API output ByteString, but that make manual test a bit harder.

See: EvaluatorImpl.createWebSocketFlow

brharrington:
Parsing directly from the ByteString is a bit faster. From a performance perspective I think we are probably better off keeping it as ByteString as much as possible. The ByteString also gives us more flexibility to use a binary JSON encoding later on for further improvement if we want to.

With this change though, the parsing can accept either giving us more flexibility to refine more over time.

PR #1149, opened 2020-03-04
cache the hash code for DataExprs

brharrington:
For streaming evaluation the DataExpr is the key in the
time grouping stage. The largest expense in the time
grouping stage was computing the hash code of the DataExpr
to aggregate the datapoints as they were flowing through.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/model/DataExpr.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/model/DataExpr.scala
@@ -22,7 +22,7 @@ import com.netflix.atlas.core.util.Math
 import com.netflix.atlas.core.util.SmallHashMap
 import com.netflix.atlas.core.util.Strings
 
-sealed trait DataExpr extends TimeSeriesExpr {
+sealed trait DataExpr extends TimeSeriesExpr with Product {
 
   def query: Query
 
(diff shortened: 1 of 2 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/model/DataExpr.scala
```diff
@@ -118,6 +118,8 @@ object DataExpr {
       val rs = consolidate(context.step, data.filter(t => query.matches(t.tags)))
       ResultSet(this, rs, context.state)
     }
+
+    override val hashCode: Int = scala.util.hashing.MurmurHash3.productHash(this)
```
jfz: 1. This can be lazy because it is not always needed
2. move it up to the DataExpr trait?

brharrington: 1. True, but lazy vals incur an additional access penalty (some more details in https://docs.scala-lang.org/sips/improved-lazy-val-initialization.html). I don't think the creation of the expression instance shouldn't be on a hot path. 

2. My general concern is that it would be evaluated before the sub-class is fully initialized.

brharrington: Re 2, it looks like it does work in this case, but not sure how comfortable I am relying on it. When scala generates the constructor for the case class it does the assignment for the main constructor fields before calling the super constructors. At the java language level that isn't allowed, but looks like it is allowed in the bytecode.

jfz: #1 Didn't know the overhead of lazy, ya I don't like that whole creation and checking in hot path either.
#2 I see, replying on that it's declared at the end of class also seems a bit worrisome, please at least add some comment so that we are aware in future

brharrington: It doesn't have to be at the end of the class as it is only based on the immutable fields passed into the constructor. My hesitation was putting it in a base class.

brharrington: I opted to move it to the base and added a comment as I think it gives us the best code clarity. If we see a regression in the future we can revisit.

jfz: sounds good, moving to trait avoids duplication and worrying about this for new subclasses.

PR #1192, opened 2020-07-30
atlas-eval: expose data rate per datasource via diagnostic messages

jfz:
Expose raw data rate (from LWC) and final data out put rate for each DataSource. 
Raw data rate is collected at TimedGroup for each DataExpr and carry it to down stream, and summarized in FinalExprEval and emit.

### DIFF
--- a/atlas-akka/src/main/scala/com/netflix/atlas/akka/DiagnosticMessage.scala
+++ b/atlas-akka/src/main/scala/com/netflix/atlas/akka/DiagnosticMessage.scala
@@ -26,7 +26,6 @@ object DiagnosticMessage {
   final val Warning: String = "warn"
   final val Error: String = "error"
   final val Close: String = "close"
-  final val Rate: String = "rate"
 
   def info(message: String): DiagnosticMessage = {
     DiagnosticMessage(Info, message, None)
@@ -42,14 +41,6 @@ object DiagnosticMessage {
 
   def error(t: Throwable): DiagnosticMessage = {
     error(s"${t.getClass.getSimpleName}: ${t.getMessage}")
-  }
-
-  def rate(dataType: String, count: Long, timestamp: Long, step: Long): DiagnosticMessage = {
-    DiagnosticMessage(
-      Rate,
-      s"dataType=$dataType,count=$count,timestamp=$timestamp,step=$step",
-      None
-    )
   }
 
   val close: DiagnosticMessage = {
--- a/atlas-akka/src/main/scala/com/netflix/atlas/akka/DiagnosticMessage.scala
+++ b/atlas-akka/src/main/scala/com/netflix/atlas/akka/DiagnosticMessage.scala
@@ -26,6 +26,7 @@ object DiagnosticMessage {
   final val Warning: String = "warn"
   final val Error: String = "error"
   final val Close: String = "close"
+  final val Rate: String = "rate"
 
   def info(message: String): DiagnosticMessage = {
     DiagnosticMessage(Info, message, None)
@@ -41,6 +42,14 @@ object DiagnosticMessage {
 
   def error(t: Throwable): DiagnosticMessage = {
     error(s"${t.getClass.getSimpleName}: ${t.getMessage}")
+  }
+
+  def rate(dataType: String, count: Long, timestamp: Long, step: Long): DiagnosticMessage = {
+    DiagnosticMessage(
+      Rate,
+      s"dataType=$dataType,count=$count,timestamp=$timestamp,step=$step",
+      None
+    )
   }
 
   val close: DiagnosticMessage = {
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
@@ -88,6 +88,8 @@ object AggrDatapoint {
     * instance have the same data expression.
     */
   trait Aggregator {
+    protected[this] var rawDatapointCounter = 0
+    def numRawDatapoints: Long = rawDatapointCounter
     def aggregate(datapoint: AggrDatapoint): Aggregator
     def datapoints: List[AggrDatapoint]
   }
@@ -101,9 +103,11 @@ object AggrDatapoint {
       extends Aggregator {
 
     private var value = init.value
+    rawDatapointCounter += 1
 
     override def aggregate(datapoint: AggrDatapoint): Aggregator = {
       value = op(value, datapoint.value)
+      rawDatapointCounter += 1
       this
     }
 
@@ -123,7 +127,9 @@ object AggrDatapoint {
 
     private def newAggregator(datapoint: AggrDatapoint): SimpleAggregator = {
       datapoint.expr match {
-        case GroupBy(af: AggregateFunction, _) => new SimpleAggregator(datapoint, af)
+        case GroupBy(af: AggregateFunction, _) =>
+          rawDatapointCounter += 1
+          new SimpleAggregator(datapoint, af)
         case _ =>
           throw new IllegalArgumentException("datapoint is not for a grouped expression")
       }
@@ -131,8 +137,10 @@ object AggrDatapoint {
 
     override def aggregate(datapoint: AggrDatapoint): Aggregator = {
       aggregators.get(datapoint.tags) match {
-        case Some(aggr) => aggr.aggregate(datapoint)
-        case None       => aggregators.put(datapoint.tags, newAggregator(datapoint))
+        case Some(aggr) =>
+          rawDatapointCounter += 1
+          aggr.aggregate(datapoint)
+        case None => aggregators.put(datapoint.tags, newAggregator(datapoint))
       }
       this
     }
@@ -150,6 +158,7 @@ object AggrDatapoint {
 
     override def aggregate(datapoint: AggrDatapoint): Aggregator = {
       values = datapoint :: values
+      rawDatapointCounter += 1
       this
     }
 
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
@@ -89,7 +89,7 @@ object AggrDatapoint {
     */
   trait Aggregator {
     protected[this] var rawDatapointCounter = 0
-    def numRawDatapoints: Long = rawDatapointCounter
+    def numRawDatapoints: Int = rawDatapointCounter
     def aggregate(datapoint: AggrDatapoint): Aggregator
     def datapoints: List[AggrDatapoint]
   }
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvalDataRateCollector.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvalDataRateCollector.scala
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2014-2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.atlas.eval.stream
+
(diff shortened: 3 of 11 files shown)


atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
```diff
@@ -149,34 +164,44 @@ private[stream] class FinalExprEval(interpreter: ExprInterpreter)
...
             val t = Instant.ofEpochMilli(timestamp)
-            k -> DiagnosticMessage.info(s"$t: ${vs.length} input datapoints for [$k]")
+            k -> DiagnosticMessage.info(s"$t: ${vs.values.length} input datapoints for [$k]")
         }
 
+        val finalDataRateMap = mutable.Map.empty[String, Long]
         // Generate the time series and diagnostic output
         val output = recipients.flatMap {
-          case (expr, ids) =>
+          case (styleExpr, ids) =>
```
jfz, line 183: just renamed for easier to follow

atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
```diff
@@ -101,9 +103,11 @@ object AggrDatapoint {
       extends Aggregator {
 
     private var value = init.value
+    rawDatapointCounter += 1
```
jfz, line 106: Need increase here because the constructor take 1 Datapoint

atlas-akka/src/main/scala/com/netflix/atlas/akka/DiagnosticMessage.scala
```diff
@@ -43,6 +44,14 @@ object DiagnosticMessage {
     error(s"${t.getClass.getSimpleName}: ${t.getMessage}")
   }
 
+  def rate(dataType: String, count: Long, timestamp: Long, step: Long): DiagnosticMessage = {
```
brharrington, line 47: This seems pretty specific to the eval library. Could we just use an info message with the right information encoded?

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
```diff
@@ -149,34 +164,44 @@ private[stream] class FinalExprEval(interpreter: ExprInterpreter)
...
-              k -> AggrDatapoint.aggregate(vs).map(_.toTimeSeries)
+              k -> AggrDatapoint.aggregate(vs.values).map(_.toTimeSeries)
           }
-        val expressionDiagnostics = groupedDatapoints.map {
+
+        val dataExprToDiagnostics = groupedDatapoints.map {
           case (k, vs) =>
             val t = Instant.ofEpochMilli(timestamp)
-            k -> DiagnosticMessage.info(s"$t: ${vs.length} input datapoints for [$k]")
+            k -> DiagnosticMessage.info(s"$t: ${vs.values.length} input datapoints for [$k]")
```
brharrington, line 177: Not sure this is still needed. I think we can refactor it to be consistent with the new messages or remove if it is redundant.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
```diff
@@ -185,14 +210,41 @@ private[stream] class FinalExprEval(interpreter: ExprInterpreter)
...
                 }
             }
         }
 
-        push(out, Source(output))
+        // Raw data rate for each DataSource
+        val rawDataRateMessages = dataSourceIdToDataExprs.map(kv =>
+          new MessageEnvelope(
+            kv._1,
+            DiagnosticMessage.rate("raw", getNumRawDatapoints(kv._2, group), group.timestamp, step)
```
brharrington, line 224: Would it make sense to output for each data expr? If it is a complex expression with many parts, then that would allow them to see which sub-expressions are contributing the most to the cost.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
```diff
@@ -149,34 +164,44 @@ private[stream] class FinalExprEval(interpreter: ExprInterpreter)
...
-                TimeSeriesMessage(expr, context, t.withLabel(expr.legend(t)))
+                TimeSeriesMessage(styleExpr, context, t.withLabel(styleExpr.legend(t)))
               }
 
-              val diagnostics = expr.expr.dataExprs.flatMap(expressionDiagnostics.get)
+              val diagnostics = styleExpr.expr.dataExprs.flatMap(dataExprToDiagnostics.get)
+
+              // Collect final output data count
+              ids.foreach(id =>
+                finalDataRateMap.get(id) match {
```
brharrington, line 200: Using `RefIntHashMap` would be more efficient here. For our current uses it is unlikely we would overflow the int, though we could create a long specialization if needed.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
```diff
@@ -185,14 +210,41 @@ private[stream] class FinalExprEval(interpreter: ExprInterpreter)
...
+        val rawDataRateMessages = dataSourceIdToDataExprs.map(kv =>
+          new MessageEnvelope(
+            kv._1,
+            DiagnosticMessage.rate("raw", getNumRawDatapoints(kv._2, group), group.timestamp, step)
+          )
+        )
+
+        // Final output data rate for each DataSource
+        val finalDataRateMessages = finalDataRateMap.map(kv => {
+          new MessageEnvelope(kv._1, DiagnosticMessage.rate("final", kv._2, group.timestamp, step))
```
brharrington, line 230: This could already be computed by the user with counting the datapoints. I think we are missing the intermediate size that the user couldn't compute. So we would have:

1. **Input size:** number of data points coming in as raw input.
2. **Intermediate size:** number of aggregate data points that are used in the eval step. This is not currently reported by this change.
3. **Output size:** number of data points for the final result.

The first two could potentially be reported per data expression. Since we expect the user to pay attention to these data points and process them, we might want to make them a standard JSON payload rather than comma separated string so they are easy to consume.

jfz: "intermediate size" looks like internal implementation detail, I wonder how is "intermediate size" useful to users,  is it actionable without changing raw input size?

brharrington: As an example consider this expression:

```
nf.app,www,:eq,
name,foo,:eq,:and,
(,id,),:by,

nf.app,www,:eq,
(,id,),:by,
1,:gt,

:mul
```

This is similar to a pattern we saw recently. In this case the second clause that only restricts by app had a high input size and intermediate size. A slight variation would be:

```
nf.app,www,:eq,
name,foo,:eq,:and,
(,id,),:by,

nf.app,www,:eq,
name,bar,:eq,:and,
(,id,),:by,
1,:gt,

:mul
```

Lets suppose that the input size is the same for both `foo` and `bar`. However, `bar` has many distinct values for `id`, but `foo` only has a few. The output size will only indicate the set that are in common between both. The difference in potential processing cost and memory usage, something we would want to support limits on, would be based on the intermediate size for the group by. Also if there is a concern about the cost, this is the difference that would tell you which clause is leading to higher cost as the input size will be the same.

atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/FinalExprEvalSuite.scala
```diff
@@ -186,19 +182,6 @@ class FinalExprEvalSuite extends AnyFunSuite {
         val ts = env.getMessage.asInstanceOf[TimeSeriesMessage]
         checkValue(ts, expectedValue)
     }
```
jfz, line 184: TBD - Will add some validation for the new data rate messages

atlas-akka/src/main/scala/com/netflix/atlas/akka/DiagnosticMessage.scala
```diff
@@ -26,6 +26,7 @@ object DiagnosticMessage {
   final val Warning: String = "warn"
   final val Error: String = "error"
   final val Close: String = "close"
+  final val Rate: String = "rate"
```
brharrington: Doesn't look like this is used anymore, it can be removed.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvalDataRateCollector.scala
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  private def increment(
+    counts: mutable.Map[String, RefIntHashMap[DataExpr]],
+    id: String,
+    dataExpr: DataExpr,
+    amount: Int
+  ): Unit = {
+    counts.get(id) match {
```
brharrington, line 80: This could be simplified to:

```scala
counts.getOrElseUpdate(id, new RefIntHashMap[DataExpr]).increment(dataExpr, amount)
```

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvalDataRateCollector.scala
```diff
@@ -0,0 +1,92 @@
...
+package com.netflix.atlas.eval.stream
+
+import com.netflix.atlas.core.model.DataExpr
+import com.netflix.atlas.core.util.RefIntHashMap
+import com.netflix.atlas.eval.model.EvalDataRate
+import com.netflix.atlas.eval.model.EvalDataSize
+
+import scala.collection.mutable
+
+class EvalDataRateCollector(timestamp: Long, step: Long) {
```
brharrington, line 25: It would be nice to have some tests for this class.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
```diff
@@ -110,9 +113,19 @@ private[stream] class FinalExprEval(interpreter: ExprInterpreter)
             }
           }
           .groupBy(_._1)
-          .map(t => t._1 -> t._2.map(_._2).toList)
+          .map(t => t._1 -> t._2.map(_._2))
           .toList
 
+        // Compute the map from DataSource id to DataExprs
+        dataSourceIdToDataExprs = mutable.Map.empty
```
brharrington, line 120: This could be refactored to keep it immutable since it is never modified later, something like:

```scala
      dataSourceIdToDataExprs = recipients
          .flatMap {
            case (styleExpr, ids) => ids.map(_ -> styleExpr.expr.dataExprs.toSet)
          }
          .foldLeft(Map.empty[String, Set[DataExpr]]) {
            case (acc, (id, dataExprs)) =>
              acc + acc.get(id).fold(id -> dataExprs) { vs => id -> (dataExprs ++ vs) }
          }
```

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
```diff
@@ -149,50 +162,65 @@ private[stream] class FinalExprEval(interpreter: ExprInterpreter)
...
+              k -> AggrDatapoint.aggregate(vs.values).map(_.toTimeSeries)
+          }
+
+        // Collect input and intermediate data size per DataSource
+        val rateCollector = new EvalDataRateCollector(timestamp, step)
+        dataSourceIdToDataExprs.foreach {
+          case (id, dataExprSet) => {
+            dataExprSet.foreach(dataExpr => {
+              group.dataExprValues.get(dataExpr).foreach { info =>
+                {
```
brharrington, line 178: This additional level of braces is unnecessary.

atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/FinalExprEvalSuite.scala
brharrington: Looks like this is still pending

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvalDataRateCollector.scala
jfz: Should be covered by FinalExprEvalSuite

atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/FinalExprEvalSuite.scala
jfz: added validations

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala
jfz: sounds good, made the change with one difference: made the init value of foldLeft mutable map to avoid creating new Map on every update

PR #1197, opened 2020-08-07
reduce allocations for asJavaMap.entrySet

brharrington:
Updates the `SmallHashMap.asJavaMap` implementation to
override `entrySet()`. The new implementation avoids
allocations for each item in the map.

**Throughput**

| Benchmark            |        Before |         After | % Delta |
|----------------------|---------------|---------------|---------|
| customEntrySet       |   8,400,934.3 |  13,480,190.3 |    60.5 |

**Allocations**

| Benchmark            |        Before |         After | % Delta |
|----------------------|---------------|---------------|---------|
| customEntrySet       |         296.0 |          24.0 |   -91.9 |

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/util/SmallHashMap.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/util/SmallHashMap.scala
@@ -442,12 +442,37 @@ final class SmallHashMap[K <: Any, V <: Any] private (val data: Array[Any], data
       override def containsKey(k: AnyRef): Boolean = self.getOrNull(k.asInstanceOf[K]) != null
 
       /**
-        * Required by AbstractMap. For this use-case we use the default scala wrapper if
-        * the entry set is needed.
+        * Overridden to avoid allocating memory for each entry when iterating over the map.
         */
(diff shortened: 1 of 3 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/util/SmallHashMap.scala
```diff
@@ -442,12 +442,37 @@ final class SmallHashMap[K <: Any, V <: Any] private (val data: Array[Any], data
...
+        * Overridden to avoid allocating memory for each entry when iterating over the map.
         */
       override def entrySet(): java.util.Set[java.util.Map.Entry[K, V]] = {
-        import scala.jdk.CollectionConverters._
-        self.asJava.entrySet()
+        new java.util.AbstractSet[java.util.Map.Entry[K, V]] {
+          override def size(): Int = self.size
+
+          override def iterator(): java.util.Iterator[java.util.Map.Entry[K, V]] = {
+            new java.util.Iterator[java.util.Map.Entry[K, V]] with java.util.Map.Entry[K, V] {
```
jfz, line 452: Great idea of using iterator itself as Entry container as well, really minimize the allocation!

PR #1205, opened 2020-09-14
allow timeout for unique operator

jfz:
Allow unique operator to emit repeated element after given time elapses. This should help keep subscriptions update to date for all lwc api instances in a cluster.

### DIFF
--- a/atlas-akka/src/main/scala/com/netflix/atlas/akka/StreamOps.scala
+++ b/atlas-akka/src/main/scala/com/netflix/atlas/akka/StreamOps.scala
@@ -382,12 +382,14 @@ object StreamOps extends StrictLogging {
 
   /**
     * Filter out repeated values in a stream. Similar to the unix `uniq` command.
+    * @param timeout Repeated value will still be emitted if elapsed time since last emit exceeds
+    *                timeout; repeated value is never emitted if timeout <= 0.
     */
-  def unique[V]: Flow[V, V, NotUsed] = {
-    Flow[V].via(new UniqueFlow[V]())
+  def unique[V](timeout: Long = 0): Flow[V, V, NotUsed] = {
+    Flow[V].via(new UniqueFlow[V](timeout))
   }
 
-  private final class UniqueFlow[V] extends GraphStage[FlowShape[V, V]] {
+  private final class UniqueFlow[V](timeout: Long) extends GraphStage[FlowShape[V, V]] {
 
     private val in = Inlet[V]("UniqueFlow.in")
     private val out = Outlet[V]("UniqueFlow.out")
@@ -398,13 +400,19 @@ object StreamOps extends StrictLogging {
 
       new GraphStageLogic(shape) with InHandler with OutHandler {
         private var previous: V = _
+        private var lastPushedAt: Long = Long.MinValue
+
+        def isExpired(): Boolean = {
+          timeout > 0 && System.currentTimeMillis() - lastPushedAt > timeout
+        }
 
         override def onPush(): Unit = {
           val v = grab(in)
-          if (v == previous) {
+          if (v == previous && !isExpired()) {
             pull(in)
           } else {
             previous = v
+            lastPushedAt = System.currentTimeMillis()
             push(out, v)
           }
         }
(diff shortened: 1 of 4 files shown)


atlas-akka/src/main/scala/com/netflix/atlas/akka/StreamOps.scala
```diff
@@ -398,13 +400,19 @@ object StreamOps extends StrictLogging {
 
       new GraphStageLogic(shape) with InHandler with OutHandler {
         private var previous: V = _
+        private var lastPushedAt: Long = Long.MinValue
+
+        def isExpired(): Boolean = {
+          timeout > 0 && System.currentTimeMillis() - lastPushedAt > timeout
```
brharrington, line 408: `Long.MinValue` is the largest negative long value, this means that `System.currentTimeMillis() - lastPushedAt` effectively adds a big long value and then overflows to become negative.

It seems like it would be logically simpler to default to `Long.MaxValue` instead of `0` and initialize `lastPushedAt` to `0`. The condition is simpler and it wouldn't rely on overflow behavior for the initial check.

atlas-akka/src/main/scala/com/netflix/atlas/akka/StreamOps.scala
```diff
@@ -382,12 +382,14 @@ object StreamOps extends StrictLogging {
 
   /**
     * Filter out repeated values in a stream. Similar to the unix `uniq` command.
+    * @param timeout Repeated value will still be emitted if elapsed time since last emit exceeds
+    *                timeout; repeated value is never emitted if timeout <= 0.
     */
-  def unique[V]: Flow[V, V, NotUsed] = {
-    Flow[V].via(new UniqueFlow[V]())
+  def unique[V](timeout: Long = 0): Flow[V, V, NotUsed] = {
```
brharrington, line 390: Consider passing in the clock, something like:

```scala
def unique[V](timeout: Long = 0, clock: Clock = Clock.SYSTEM): Flow[V, V, NotUsed]
```

This makes it easier to test by using a mock implementation of the clock to control the timing.

atlas-akka/src/test/scala/com/netflix/atlas/akka/StreamOpsSuite.scala
```diff
@@ -252,7 +252,7 @@ class StreamOpsSuite extends AnyFunSuite {
 
   test("unique") {
```
brharrington, line 253: Can you add a test for the timeout case?

atlas-akka/src/test/scala/com/netflix/atlas/akka/StreamOpsSuite.scala
```diff
@@ -252,10 +252,29 @@ class StreamOpsSuite extends AnyFunSuite {
...
-      .via(StreamOps.unique)
+      .via(StreamOps.unique())
       .runWith(Sink.seq[Int])
     val vs = Await.result(future, Duration.Inf)
     // Only consecutive repeated values are filtered out, so the final 1 should get repeated
     assert(vs === List(1, 2, 3, 4, 5, 6, 7, 1))
   }
+
+  test("unique timeout") {
+    println(Long.MaxValue == Long.MaxValue)
```
brharrington, line 263: Is this needed?

atlas-akka/src/main/scala/com/netflix/atlas/akka/StreamOps.scala
```diff
@@ -382,12 +383,17 @@ object StreamOps extends StrictLogging {
 
   /**
     * Filter out repeated values in a stream. Similar to the unix `uniq` command.
+    *
+    * @param timeout
+    *     Repeated value will still be emitted if elapsed time since last emit exceeds
```
brharrington, line 388: Consider mentioning that unit is milliseconds.

brharrington (approved):
I went ahead and pushed changes for the two minor comments.

jfz:
> I went ahead and pushed changes for the two minor comments. Just waiting to confirm that CI build is successful.
Thanks, looks good to me!

PR #1234, opened 2020-12-14
add cardinality estimator, based on datasketches cpc

jfz:
add cardinality estimator, based on datasketches cpc


brharrington (approved):
I added some comments, but otherwise looks good to me.

PR #1237, opened 2021-01-12
cardinality stats tracking by app, name and tags

jfz:
cardinality stats tracking by app, name and tags


atlas-core/src/main/scala/com/netflix/atlas/core/rollup/CardinalityStats.scala
```diff
@@ -0,0 +1,255 @@
...
+      .map(entry => (entry.getKey, entry.getValue.cardinality))
+  }
+
+  /**
+    * Iterator for app, name and cardinality tuples.
+    *
+    * @return
+    *     iterator for app, name and cardinality tuples
+    */
+  def appNameCardinalityIterator(): Iterator[(String, String, Int)] = {
```
brharrington, line 74: How do you envision these stat queries getting used?

One thing we would need is a way to take a `DataExpr` and verify if it can be satisfied without using any of the dimensions that would be rolled up. For more complex use-cases it would be helpful if there was a way to see partial applicability, e.g., suppose we have a query across the fleet:

```
name,ipc.server.call,:eq,statistic,count,:eq,:and,:sum,(,nf.cluster,),:by
```

However, `nf.cluster` is suppressed for a small number of apps (a, b, and c), then we could partially use the rollup view and partially use the more expensive fallback:

```
Memory:
name,ipc.server.call,:eq,statistic,count,:eq,:and,nf.app,(,a,b,c,),:in,:not,:and,:sum,(,nf.cluster,),:by

Disk:
name,ipc.server.call,:eq,statistic,count,:eq,:and,nf.app,(,a,b,c,),:in,:and,:sum,(,nf.cluster,),:by

Final:
merge memory and disk views
```

atlas-core/src/main/scala/com/netflix/atlas/core/rollup/CardinalityStats.scala
```diff
@@ -0,0 +1,255 @@
...
+    val appStats = appToAppStats.get(app)
+    if (appStats == null) {
+      0
+    } else {
+      appStats.cardinalityByName(name)
+    }
+  }
+}
+
+private class AppStats {
```
brharrington, line 163: I wonder if these could be made a little bit more generic so it would work like a tree. This would give us flexibility to configure a path, for right now it would be `nf.app > name`. Would make it easy to change or test other possible variants using historical data.

PR #1239, opened 2021-01-19
add mapping for elb processed bytes

jfz:
add mapping for elb processed bytes

### DIFF
--- a/atlas-poller-cloudwatch/src/main/resources/elb.conf
+++ b/atlas-poller-cloudwatch/src/main/resources/elb.conf
@@ -14,6 +14,11 @@ atlas {
 
       metrics = [
         {
+          name = "ProcessedBytes"
+          alias = "aws.elb.processedBytes"
+          conversion = "sum,rate"
+        },
+        {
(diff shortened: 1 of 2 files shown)


atlas-poller-cloudwatch/src/main/resources/elb.conf
```diff
@@ -13,6 +13,11 @@ atlas {
       ]
 
       metrics = [
+        {
+          name = "ProcessedBytes"
```
brharrington, line 17: I see this for ALB/NLB, but not classic load balancers ([docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-cloudwatch-metrics.html)). There is an `EstimatedProcessedBytes` metric listed in the docs, however, I don't see it in the console for our load balancers.

jfz: EstimatedProcessedBytes is actually available, and add as a different category because it has different dimension with other elb metrics.

brharrington (approved):

PR #1244, opened 2021-01-28
binary ops with a subset of group by keys

brharrington:
This change add support for binary operations where both
sides are grouped and the keys for one side is a subset
of the other. Before the keys would have to match exactly.
When one side has a subset, the matching for entries will
be performed using the smaller set of keys.

For example, suppose that dimensions `a` and `b` each have
two values:

```
Expr: q1,(,a,b,),:by, q2,(,a,),:by, :add

Result Set:
      q1:a1b1 + q2:a1
      q1:a1b2 + q2:a1
      q1:a2b1 + q2:a2
      q1:a2b2 + q2:a2
```

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/model/DataExpr.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/model/DataExpr.scala
@@ -34,9 +34,7 @@ sealed trait DataExpr extends TimeSeriesExpr with Product {
 
   def isGrouped: Boolean = false
 
-  def groupByKey(tags: Map[String, String]): Option[String] = {
-    Some(TaggedItem.computeId(tags).toString)
-  }
+  def groupByKey(tags: Map[String, String]): Option[String] = None
 
   def finalGrouping: List[String] = Nil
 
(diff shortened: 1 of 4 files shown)


jfz:
Exact match sounds reasonable, sub set match result can be confusing for users, what kind use cases benefit from this?

brharrington:
An example use-case is computing the percentage of different types of failures for a set of clusters. The numerator is the number of failures by cluster and status. To get a percentage relative to a given cluster, the lines for a given cluster need to be divided by the total number of requests for that cluster.

```
name,ipc.server.call,:eq,statistic,count,:eq,:and,ipc.result,failure,:eq,:and,(,nf.cluster,ipc.status,),:by,
name,ipc.server.call,:eq,statistic,count,:eq,:and,(,nf.cluster,),:by,
:div,100,:mul
```

I'm not sure it creates more confusion. I have had some power users kind of assume it would work in the past and were surprised it did not. However, there is a bit of selection bias there, so may not generalize well to users that are less familiar with Atlas. We already support it where one side is grouped and the other isn't:

```
name,ipc.server.call,:eq,statistic,count,:eq,:and,ipc.result,failure,:eq,:and,(,ipc.status,),:by,
name,ipc.server.call,:eq,statistic,count,:eq,:and,:sum,
:div,100,:mul
```

This change makes it a bit more regular as that can be thought of as one side having an empty set for the grouping keys. We are no longer special casing that one example of a subset.

jfz (approved):

Issue #1224, opened 2020-12-03
proposal for top/bottom-K operators


brharrington:
## Basic Top/Bottom-K

The proposed syntax would be similar to the existing [:stat](https://github.com/Netflix/atlas/wiki/filter-stat) commands. Example:

```
group-by-expression,S,N,:topk
group-by-expression,S,N,:bottomk
```

The `topk` and `bottomk` names match the similar names for the same sort of operation used by [Prometheus](https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators). `S` would be one of the available summary statistics: `avg`, `max`, `min`, `sum`, `last`, `count`. `N` is a positive integer indicating the number of time series to include in the result set. There will be a limit on the maximum size of `N` to bound resources. Exact value not yet decided, but probably ~100 for default config.

For consistency with `:stat` there would also be helpers like `:topk-{statistic}`. Examples:

```
name,sps,:eq,(,nf.node,),:by,max,5,:topk,
name,sps,:eq,(,nf.node,),:by,5,:topk-max,

name,sps,:eq,(,nf.node,),:by,avg,5,:bottomk,
name,sps,:eq,(,nf.node,),:by,5,:bottomk-avg,
```

## Include Others

It is desirable in some cases to have other categories that are not part of the top/bottom-K included as an aggregated entry. The aggregation function used for those entries will be the same as that used for the grouping operation. Example:

```
name,sps,:eq,:sum,(,nf.node,),:by,max,5,:topk-others,
```

This would return up to 6 results. The 5 nodes that have the max SPS individually and a sixth result that would be the sum of all other nodes.

## Math

The top/bottom-K operators can be used anywhere on a grouped expression including after math expressions combining multiple sources. This allows it to be used on the result of computed aggregates like averages, percent errors, percentile approximations, etc. Example:

```
statistic,totalTime,:eq,:sum,
statistic,count,:eq,:sum,:div,
(,nf.node,),:by,
name,latency,:eq,:cq,
5,:topk-max
```

## Streaming

Top/bottom-K operators would not be supported for streaming execution.

PR #1245, opened 2021-01-28
add topk/bottomk operators

brharrington:
Add filtering operators to collect the top/bottom `K`
time series based on a summary statistic. This is the
first part of #1224. A follow up PR will add the
variant that aggregates the others category.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/model/FilterVocabulary.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/model/FilterVocabulary.scala
@@ -23,6 +23,7 @@ import com.netflix.atlas.core.stacklang.Word
 object FilterVocabulary extends Vocabulary {
 
   import com.netflix.atlas.core.model.ModelExtractors._
+  import com.netflix.atlas.core.stacklang.Extractors._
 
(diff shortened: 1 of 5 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/model/FilterVocabulary.scala
```diff
@@ -223,4 +227,60 @@ object FilterVocabulary extends Vocabulary {
...
+    override def signature: String = "TimeSeriesExpr stat:String k:Int -- FilterExpr"
+
+    override def summary: String =
+      """
+        |Limit the output to the `K` time series with the highest values for the specified
+        |summary statistic.
+      """.stripMargin.trim
+
+    override def examples: List[String] =
+      List("name,sps,:eq,:sum,(,nf.cluster,),:by,max,5")
```
jfz, line 256: Missing :topk at end?

brharrington: The operator itself gets automatically added to all of the examples.

jfz (approved):

PR #1248, opened 2021-02-02
add topk/bottomk that aggregates others

### LINKED ISSUE #1224: proposal for top/bottom-K operators

brharrington:
Fixes #1224.

Adds variants of topk/bottomk that in addition to computing
the highest priority values will also return an aggregate
time series that includes all of the other time series that
were not high enough priority. This can be useful to see
what proportion of the overall volume is represented by
the highest priority time series. The operators have the
same signature as `:topk` and `:bottomk`, but the
aggregation to use is specified in the operator name.
There are four aggregates supported:

- min
- max
- sum
- avg

Count was not included as an aggregate because it could be
ambiguous and does not seem useful.

As part of this work there was some refactoring of the
existing aggregation operations to make the count case
more consistent.


brharrington:
Note, this PR depends on #1247 and will need to be updated once that is merged.

brharrington:
Regarding count, the reason it was considered problematic is that it could mean the aggregate count across others or the count of how many others were present. For example:

```
Expr:
    Q,:count,(,foo,),:by

Output:
    foo1: 1
    foo2: 2
    foo3: 3 (means 3 time series had values for foo3)

Expr with priority operator:
    Q,:count,(,foo,),:by,max,1,:bottomk-others-count

Option 1:
    foo1:       1 (highest priority)
    --others--: 5 (cumulative number of input time series contributing to others)

Option 2:
    foo1:       1 (highest priority)
    --others--: 2 (number of time series that were not high enough priority)
```

jfz (approved):

PR #1250, opened 2021-02-03
add mapping for 2 rds metrics

jfz:
add mapping for 2 rds metrics

### DIFF
--- a/atlas-poller-cloudwatch/src/main/resources/rds.conf
+++ b/atlas-poller-cloudwatch/src/main/resources/rds.conf
@@ -100,8 +100,14 @@ atlas {
         },
         {
           name = "AuroraVolumeBytesLeftTotal"
-          alias = "aws.rds.auroraVolumeBytesLeftTotal"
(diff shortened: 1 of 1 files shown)


atlas-poller-cloudwatch/src/main/resources/rds.conf
```diff
@@ -98,6 +98,16 @@ atlas {
           alias = "aws.rds.binLogReplicaLag"
           conversion = "max"
         },
+        {
+          name = "AuroraVolumeBytesLeftTotal"
+          alias = "aws.rds.auroraVolumeBytesLeftTotal"
```
brharrington, line 103: Should this be modeled with an engine tag similar to other metrics in case we need to map in the same concept from some other engine in the future? For example:

https://github.com/Netflix/atlas/blob/master/atlas-poller-cloudwatch/src/main/resources/rds.conf#L75

brharrington (approved):

PR #1253, opened 2021-02-05
fix dimension for rds metric: VolumeBytesUsed

jfz:
Use the right dimension "DbClusterIdentifier" for rds metric "VolumeBytesUsed".

### DIFF
--- a/atlas-poller-cloudwatch/src/main/resources/rds.conf
+++ b/atlas-poller-cloudwatch/src/main/resources/rds.conf
@@ -110,11 +110,6 @@ atlas {
           ]
         },
         {
-          name = "VolumeBytesUsed"
-          alias = "aws.rds.volumeBytesUsed"
-          conversion = "max"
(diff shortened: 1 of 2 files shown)


atlas-poller-cloudwatch/src/main/resources/rds.conf
```diff
@@ -426,5 +421,23 @@ atlas {
...
       ]
     }
+
+    rds-cluster = {
+      namespace = "AWS/RDS"
+      period = 1m
+      timeout = 15m
+
+      dimensions = [
+        "DbClusterIdentifier"
```
brharrington, line 431: Might want to map this to an aws prefixed tag key like we do for the instance id.

https://github.com/Netflix/atlas/blob/master/atlas-poller-cloudwatch/src/main/resources/reference.conf#L198

brharrington (approved):

PR #1276, opened 2021-03-17
lwc: helper for generating data

brharrington:
Adds a helper that can be used to generate arbitrary
amounts of data for an expression by changing the scheme
to `synthetic`.


atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/Evaluator.java
```diff
@@ -302,7 +302,8 @@ public String getUri() {
     public boolean isLocal() {
       return uri.startsWith("/")
           || uri.startsWith("file:")
-          || uri.startsWith("resource:");
+          || uri.startsWith("resource:")
+          || uri.startsWith("synthetic:");
```
jfz, line 306: I like the name!

jfz (approved):

PR #1280, opened 2021-03-19
auto rollup or drop tags based on cardinality limits

jfz:
Automatically rollup or drop tags based on cardinality limit configuration, also provide an api to check if a query can be served based on the limiter stats. 

It tracks cardinality by a list of pre-defined keys and associated limits, and build out a tree structure level by level based on that, also tracks cardinality for other tag keys at leaf level.

Given below example configuration:
`atlas.auto-rollup = {
  prefix = `[`
    {
      key = "app",
      value-limit = 20,
      total-limit = 1000,
    },
    {
      key = "name",
      value-limit = 30,
      total-limit = 50,
    }
  ],
  tag-value-limit = 40
}`

- Level 1: track cardinality by "app", max number of apps allowed is 20, max number of tags across all apps is 1000; drop new apps if either limit is reached.

- Level 2: for a specific app, track cardinality by "name", max number of names is 30, max number of tags across all names is 50; drop new names if limit reached; drop new names if either limit is reached.

- Level 3: for a specific app and name, tracks number of unique values per non prefix tag keys, max number of values is 40; rollup if the limit is reached.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/limiter/CardinalityLimiter.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/limiter/CardinalityLimiter.scala
@@ -1,0 +1,334 @@
+/*
+ * Copyright 2014-2021 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.atlas.core.limiter
+
+import com.netflix.atlas.core.model.Query
+import com.netflix.atlas.core.model.TaggedItem
+import com.netflix.atlas.core.limiter.CardinalityLimiter._
+import com.netflix.atlas.core.util.BoundedPriorityBuffer
+import com.netflix.atlas.core.util.CardinalityEstimator
+import com.typesafe.config.Config
+
+import java.util.concurrent.ConcurrentHashMap
+import scala.jdk.CollectionConverters._
+
+/**
+  * A structure that tracks cardinality based on tag keys, and drop or rollup tags based on given
+  * configuration.
+  */
+abstract class CardinalityLimiter(val limiterConfig: LimiterConfig) {
+
+  protected val totalEstimator: CardinalityEstimator = CardinalityEstimator.newSyncEstimator()
+
+  /**
+    * Update cardinality stats for the given tags.
+    * @param tags
+    *   tags to track
+    * @param id
+    *   id that represents the tags
+    * @return
+    *   - null if it should be dropped
+    *   - updated tags if tag roll up happened
+    *   - same object with input tags otherwise
+    */
+  def update(tags: Map[String, String], id: AnyRef): Map[String, String]
+
+  /**
+    * Default to [[TaggedItem.computeId]] to compute id.
+    */
+  def update(tags: Map[String, String]): Map[String, String] = {
+    update(tags, TaggedItem.computeId(tags))
+  }
+
+  /** Get total cardinality for total tags ever seen. */
+  def cardinality: Long = totalEstimator.cardinality
+
+  /**
+    * Get cardinality for a given list of values of prefix keys.
+    */
+  def cardinalityBy(values: List[String]): Long
+
+  /**
+    * Get cardinality for a given list of values of prefix keys and a specific tag key.
+    */
+  def cardinalityBy(values: List[String], tagKey: String): Long
+
+  /**
+    * Check if a query can be served based on the cardinality stats and defined limits.
+    */
+  def canServe(query: Query): Boolean = canServe(QueryInfo(query, limiterConfig))
+
+  private[limiter] def canServe(queryInfo: QueryInfo): Boolean
+
+  /**
+    * A convenient way get topk keys by cardinality at all levels, mainly used for debug/inspect.
+    */
+  def topk(k: Int): AnyRef
+}
+
+/**
+  * Non-leaf node of the tree, tracks stats by prefix keys.
+  */
+private[limiter] class CardinalityLimiterInner(
+  override val limiterConfig: LimiterConfig,
+  val level: Int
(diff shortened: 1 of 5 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/limiter/CardinalityLimiter.scala
```diff
@@ -0,0 +1,334 @@
...
+  val RollupValue = "_auto-rollup_"
+  val MissingKey = "_missing-key_"
+}
+
+case class QueryInfo(query: Query, limiterConfig: LimiterConfig) {
+  private var _queryKeys: Set[String] = _
+  val prefixValues: Array[String] = genSearchPath()
+
+  // Not needed if drop found early in search path, so generate lazily
+  def queryKeys: Set[String] = {
```
brharrington, line 311: Could you just declare it a `lazy val`?

atlas-core/src/main/scala/com/netflix/atlas/core/util/BoundedPriorityBuffer.scala
```diff
@@ -81,4 +81,13 @@ class BoundedPriorityBuffer[T <: AnyRef](maxSize: Int, comparator: Comparator[T]
     }
     builder.result()
   }
+
+  /** Return a list containing all of the items in the buffer - preserving order by priority. */
```
brharrington, line 85: Not sure if there is a better name (`drainToOrderedList` similar to `BlockingQueue.drainTo`?), I think we should at lease update the scaladoc comment to indicate that this will empty out the buffer and doesn't just return a copy of the elements.

atlas-core/src/main/scala/com/netflix/atlas/core/util/CardinalityEstimator.scala
```diff
@@ -29,13 +31,16 @@ trait CardinalityEstimator {
...
+
+  override def toString: String = cardinality.toString
 }
 
 object CardinalityEstimator {
 
   /**
-    * Create a new estimator instance using the [CPC] algorithm.
+    * Create a new estimator instance using the [CPC] algorithm. This created estimator is NOT
+    * thread safe, use {@link newSyncEstimator} to create a thread-safe estimator.
```
brharrington, line 43: Do we need both? Can we just make `CpcEstimator` thread safe?

atlas-core/src/main/scala/com/netflix/atlas/core/util/CardinalityEstimator.scala
```diff
@@ -48,13 +53,30 @@ object CardinalityEstimator {
...
 
-    override def cardinality: Int = sketch.getEstimate.toInt
+    override def cardinality: Long = sketch.getEstimate.longValue()
+  }
+
+  private class SyncCpcEstimator(val lgK: Int) extends CardinalityEstimator {
+    private val sketch = new CpcSketch(lgK)
+    private val _cardinality = new AtomicLong()
+
+    override def update(obj: AnyRef): Unit = {
```
brharrington, line 75: This is only thread safe for reads, correct?

brharrington (commented):
Looks good, added some minor notes and questions

atlas-core/src/main/scala/com/netflix/atlas/core/limiter/CardinalityLimiter.scala
```diff
@@ -0,0 +1,334 @@
...
+  override def update(
+    tags: Map[String, String],
+    id: AnyRef
+  ): Map[String, String] = {
+    val key = limiterConfig.prefixKeys(level)
+    val value = tags.getOrElse(key, MissingKey)
+    val conf = limiterConfig.getPrefixConfig(level)
+
+    def reachLimit: Boolean = {
+      conf.totalLimit > 0 && (cardinality >= conf.totalLimit || children.size() >= conf.valueLimit)
```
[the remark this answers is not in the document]
jfz: 0 means no limit because reachLimit return false, can be useful for perf test or if we don't want to apply limit at root level.

atlas-core/src/main/scala/com/netflix/atlas/core/limiter/CardinalityLimiter.scala
```diff
@@ -0,0 +1,334 @@
...
+    * Check if a query can be served based on the cardinality stats and defined limits.
+    */
+  def canServe(query: Query): Boolean = canServe(QueryInfo(query, limiterConfig))
+
+  private[limiter] def canServe(queryInfo: QueryInfo): Boolean
+
+  /**
+    * A convenient way get topk keys by cardinality at all levels, mainly used for debug/inspect.
+    */
+  def topk(k: Int): AnyRef
```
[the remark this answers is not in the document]
jfz: introduced a new type: CardinalityStats

brharrington (approved):

atlas-core/src/main/scala/com/netflix/atlas/core/limiter/CardinalityLimiter.scala
```diff
@@ -0,0 +1,334 @@
...
+
+  override def update(
+    tags: Map[String, String],
+    id: AnyRef
+  ): Map[String, String] = {
+    val key = limiterConfig.prefixKeys(level)
+    val value = tags.getOrElse(key, MissingKey)
+    val conf = limiterConfig.getPrefixConfig(level)
+
+    def reachLimit: Boolean = {
```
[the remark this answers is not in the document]
jfz: Extracted it to class level without adding extra member field because number of instances is likely to be high for this class.

atlas-core/src/main/scala/com/netflix/atlas/core/limiter/CardinalityLimiter.scala
```diff
@@ -0,0 +1,334 @@
...
+    }
+  }
+
+  private def rollup(tags: Map[String, String], rollupKeys: Seq[String]): Map[String, String] = {
+    tags.map {
+      case (k, v) =>
+        if (rollupKeys.contains(k))
+          (k, RollupValue)
+        else
+          (k, v)
```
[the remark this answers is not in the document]
jfz: Changed to use tuple for iteration instead since it's in hotpath

atlas-core/src/main/scala/com/netflix/atlas/core/util/CardinalityEstimator.scala
jfz: updated, so far we need only one.

atlas-core/src/main/scala/com/netflix/atlas/core/util/CardinalityEstimator.scala
jfz: Right, added some comments for this.

PR #1288, opened 2021-04-01
rearrange colors for light/dark 24 palettes

### LINKED ISSUE #1287: update dark mode palette ordering
- First two colors are red and green making it hard to distinguish for folks with certain vision types.
- Some interpret red as an errors and would prefer if that wasn't the default color for single line charts.

/cc @robzienert

brharrington:
Rearrance the colors to visually follow the order of
the armytage palette. This should address some issues
raised with the light24 palette used by default with
the dark theme.

Fixes #1287


brharrington:
Colors in the armytage, dark24, light24 compared:

palettes

PR #1291, opened 2021-04-15
add mapping for Neptune metrics

jfz:


atlas-poller-cloudwatch/src/main/resources/reference.conf
```diff
@@ -306,6 +306,9 @@ atlas {
       "lambda",
       "lambda-fn-res",
       "nat-gateway",
+      "neptune",
```
brharrington, line 309: Neptune config needs to be included below.

atlas-poller-cloudwatch/src/main/resources/neptune.conf
```diff
@@ -0,0 +1,236 @@
+
+atlas {
+  cloudwatch {
+
+    // https://docs.aws.amazon.com/neptune/latest/userguide/cw-metrics.html
+    neptune = {
+      namespace = "AWS/neptune"
+      period = 1m
```
brharrington, line 8: The docs say everything is 5m, does that not match what we actually see?

atlas-poller-cloudwatch/src/main/resources/neptune.conf
```diff
@@ -0,0 +1,236 @@
...
+            }
+          ]
+        }
+      ]
+    }
+
+    # Metrics published at cluster level, every 5 minutes
+    neptune-cluster-5m = {
+      namespace = "AWS/neptune"
+      period = 1m
```
brharrington, line 177: Why isn't it using a 5m period?

atlas-poller-cloudwatch/src/main/resources/neptune.conf
```diff
@@ -0,0 +1,236 @@
...
+            }
+          ]
+        }
+      ]
+    }
+
+    # Metrics published at cluster level, once per day
+    neptune-cluster-1d = {
+      namespace = "AWS/neptune"
+      period = 1m
```
brharrington, line 219: Should this use a 1d period?

atlas-poller-cloudwatch/src/main/resources/neptune.conf
```diff
@@ -0,0 +1,236 @@
+
+atlas {
+  cloudwatch {
+
+    // https://docs.aws.amazon.com/neptune/latest/userguide/cw-metrics.html
+    neptune = {
+      namespace = "AWS/neptune"
```
brharrington, line 7: Not sure if it is case sensitive, but looks like it should be `AWS/Neptune`

atlas-poller-cloudwatch/src/main/resources/neptune.conf
```diff
@@ -0,0 +1,236 @@
...
+        },
+        {
+          name = "ClusterReplicaLagMinimum"
+          alias = "aws.neptune.replicaLagMinimum"
+          conversion = "min"
+        },
+        {
+          name = "GremlinRequestsPerSec"
+          alias = "aws.neptune.requests"
+          conversion = "max"
```
brharrington, line 54: Should this be `sum,rate`?

atlas-poller-cloudwatch/src/main/resources/neptune.conf
```diff
@@ -0,0 +1,236 @@
...
+            {
+              key = "id"
+              value = "out"
+            }
+          ]
+        },
+        {
+          name = "NumTxCommitted"
+          alias = "aws.neptune.transactions"
+          conversion = "max"
```
brharrington, line 119: Docs claim this is a rate per second, should it be `sum,rate`?

atlas-poller-cloudwatch/src/main/resources/neptune.conf
```diff
@@ -0,0 +1,236 @@
...
+      timeout = 15m
+
+      dimensions = [
+        "DbClusterIdentifier"
+      ]
+
+      metrics = [
+        {
+          name = "VolumeBytesUsed"
+          alias = "aws.neptune.volumneBytesUsed"
```
brharrington, line 188: `s/volumneBytesUsed/volumeBytesUsed/`

atlas-poller-cloudwatch/src/main/resources/neptune.conf
jfz: yes, most of them actually are being published minutely

brharrington: Not sure how neptune data is reported and if this would apply, but for EC2 data at the auto-scaling group level can sometimes appear to be minute level even though the samples from each node making up the ASG are only reported every 5m. This is because each node reports independently and may not be aligned to 5m boundaries. So some nodes report at minutes 0, 5, 10, 15, ... and some other nodes report at 1, 6, 11, 16, ... and so on.

jfz: I confirmed that it's minutely for individual instance too.

atlas-poller-cloudwatch/src/main/resources/neptune.conf
jfz: I was worried about big period making data stable, but I think I can avoid it by setting "end-period-offset" to 0 in such case.

brharrington (approved):

PR #1295, opened 2021-04-20
log the exception when events are dropped due to queue failure

lavanyachennupati:
Log the exception when events are dropped due to stream failure before or during call to the stream.

This gives insight into what caused the stream failure.

### DIFF
--- a/atlas-akka/src/main/scala/com/netflix/atlas/akka/StreamOps.scala
+++ b/atlas-akka/src/main/scala/com/netflix/atlas/akka/StreamOps.scala
@@ -124,8 +124,12 @@ object StreamOps extends StrictLogging {
         case Success(QueueOfferResult.Enqueued)    => enqueued.increment()
         case Success(QueueOfferResult.Dropped)     => dropped.increment()
         case Success(QueueOfferResult.QueueClosed) => closed.increment()
-        case Success(QueueOfferResult.Failure(_))  => failed.increment()
-        case Failure(t)                            => completed.increment()
+        case Success(QueueOfferResult.Failure(t)) =>
+          logger.warn(s"stream failed before or during call to the stream", t)
+          failed.increment()
+        case Failure(t) =>
+          logger.warn(s"stream completed with failure", t)
+          completed.increment()
       }
     }
   }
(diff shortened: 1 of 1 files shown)


atlas-akka/src/main/scala/com/netflix/atlas/akka/StreamOps.scala
```diff
@@ -124,8 +124,12 @@ object StreamOps extends StrictLogging {
         case Success(QueueOfferResult.Enqueued)    => enqueued.increment()
         case Success(QueueOfferResult.Dropped)     => dropped.increment()
         case Success(QueueOfferResult.QueueClosed) => closed.increment()
-        case Success(QueueOfferResult.Failure(_))  => failed.increment()
-        case Failure(t)                            => completed.increment()
+        case Success(QueueOfferResult.Failure(t)) =>
+          logger.warn(s"stream failed before or during call to the stream", t)
```
jfz, line 128: It can add a lot of overhead if it happens frequently for whatever reason, I'd suggest putting it at DEBUG level and enable on demand for troubleshooting.

lavanyachennupati: Sure! We should be able to add threshold based alerts to identify if this were to happen often and enable debug logging on demand.

lavanyachennupati: My initial thinking for adding `logger.warn` was similar to how one could add `StreamOps.supervisionStrategy(registry))` to get insight into exceptions that logs the exception at `warn` level. Perhaps the use case for adding a `supervisionStrategy` is on demand :)

brharrington: Is there a reason you don't just enable logging on the stream?

https://doc.akka.io/docs/akka/current/stream/stream-error.html#logging-errors

It seems like that would be a lot less noisy (once per stream vs once per insert) and would keep the logging calls out of the hot path.

brharrington: Note, `StreamOps.supervisionStrategy` was meant for use with `ActorMaterializer` which would allow for an easy way to provide default supervision strategy for all streams using that materializer ([old docs](https://doc.akka.io/docs/akka/2.4/scala/stream/stream-error.html)). This was convenient for many of our use-cases, however, that was deprecated in 2.6.0. You can still use that if you want, but would need to set the attributes for a particular stream explicitly:

https://doc.akka.io/docs/akka/current/stream/stream-error.html#supervision-strategies

lavanyachennupati: Enabling logging on the stream helped us get insights into the exceptions we were looking for. Thanks. Will close out this pr.

PR #1296, opened 2021-04-23
add mapping for AWS neptune metrics with dimension of cluster and role.

jfz:
AWS Neptune metrics by "cluster and role" are published as separate time series, need to add them in addition to existing ones that are by instance or cluster only.


jfz:
Metrics are published either with dimension of either "DBClusterIdentifier + Role" or "DBInstanceIdentifier".

brharrington (commented):
It is unfortunate they have disjoint dimensions for the same metrics. The downside to mapping in both is that if the user just queries based on the name, then they would get duplication since it will match both sets.

Are both really needed? Maybe it would be better to just map in the cluster/role set to minimize the risk of confusion.

jfz:
> It is unfortunate they have disjoint dimensions for the same metrics. The downside to mapping in both is that if the user just queries based on the name, then they would get duplication since it will match both sets.
> 
> Are both really needed? Maybe it would be better to just map in the cluster/role set to minimize the risk of confusion.

Users like to use both, I think it wouldn't be too confusing since the dimension sets do not overlap.

brharrington:
> I think it wouldn't be too confusing since the dimension sets do not overlap.

The issue is more that you have to select both name and one of those other dimensions to get a meaningful aggregate. Typically we recommend structuring data so that it can be meaningfully aggregated if just the name is selected and the dimensions can be used to drill down into the data.

PR #1306, opened 2021-05-19
allow multiple subscribers for eval processor

jfz:
The processor created by Flow.toProcessor only allows single subscriber, but some use cases need multiple subscribers.

### DIFF
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
@@ -15,14 +15,6 @@
  */
 package com.netflix.atlas.eval.stream
 
-import java.nio.file.OpenOption
-import java.nio.file.Path
-import java.nio.file.StandardOpenOption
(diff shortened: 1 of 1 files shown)


atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
```diff
@@ -167,7 +168,23 @@ private[stream] abstract class EvaluatorImpl(
   }
 
   protected def createStreamsProcessorImpl(): Processor[DataSources, MessageEnvelope] = {
-    createStreamsFlow.toProcessor.run()
+    Source
+      .asSubscriber[DataSources]
+      .via(createStreamsFlow)
+      .toMat(Sink.asPublisher[MessageEnvelope](true))(
```
jfz, line 174: setting fanout to true, this is the difference with Flow.toProcess,

brharrington (commented):
After further debugging we found the issue was due to Akka's subscription timeout. Avoiding the delays between the processor being created and something subscribing to it cleared up the issue.

PR #1309, opened 2021-05-29
Attempt to fix dropped-old in LWC client API

yingwuzhao:


brharrington:
The logic as is appears to be correct, the data needs to be grouped by the step and fed through so that all data for a corresponding interval is available for evaluation. Grouping by the id is not correct there.

PR #1377, opened 2021-11-02
Common-group-by operator

### LINKED ISSUE #1354: Common query type construct for group by
For some use-cases it might be helpful to have a common query type operator that would allow someone to append a rewrite that would include additional keys to the group by clause. Example:

```
name,foo,:eq,(,a,b,),:by,
name,bar,:eq,(,b,),:by,:div,
name,baz,:eq,:mul,
(,c,d,),:cg
```

The `:cg` is a hypothetical common grouping operator. The expression above would be equivalent to:

```
name,foo,:eq,(,a,b,c,d,),:by,
name,bar,:eq,(,b,c,d,),:by,:div,
name,baz,:eq,(,c,d,),:by,:mul
```

/cc @ruchirj

dmuino:
This adds a common-group-by operator (`:cg`) that includes additional
keys in group-by clauses.

```
name,foo,:eq,(,a,b,),:by,
name,bar,:eq,(,b,),:by,:div,
name,baz,:eq,:mul,
(,c,d,),:cg
```

Gets rewritten to:

```
name,foo,:eq,(,a,b,c,d,),:by,
name,bar,:eq,(,b,c,d,),:by,:div,
name,baz,:eq,(,c,d,),:by,:mul
```

Fixes #1354

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/model/DataVocabulary.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/model/DataVocabulary.scala
@@ -35,6 +35,7 @@ object DataVocabulary extends Vocabulary {
     Min,
     Max,
     GroupBy,
+    CommonGroupBy,
     Offset,
     CfAvg,
     CfSum,
@@ -160,9 +161,7 @@ object DataVocabulary extends Vocabulary {
       """.stripMargin.trim
   }
 
-  case object GroupBy extends SimpleWord {
-
-    override def name: String = "by"
+  trait DataGroupBy extends SimpleWord {
 
     protected def matcher: PartialFunction[List[Any], Boolean] = {
       case (_: List[_]) :: (_: Query) :: _             => true
@@ -190,6 +189,15 @@ object DataVocabulary extends Vocabulary {
         "name,sps,:eq,:max,(,nf.cluster,)",
         "name,sps,:eq,nf.cluster,nccp-silverlight,:eq,:and,(,nf.asg,nf.zone,)"
       )
+  }
+
+  case object GroupBy extends DataGroupBy {
+    override def name: String = "by"
+  }
+
+  // In this context :cg is just an alias to :by
+  case object CommonGroupBy extends DataGroupBy {
+    override def name: String = "cg"
   }
 
   case object Offset extends SimpleWord {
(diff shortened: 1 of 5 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/model/DataVocabulary.scala
```diff
@@ -192,6 +191,15 @@ object DataVocabulary extends Vocabulary {
       )
   }
 
+  case object GroupBy extends DataGroupBy {
+    override def name: String = "by"
+  }
+
+  // In this context :cg is just an alias to :by
+  case object CommonGroupBy extends DataGroupBy {
```
brharrington, line 199: Why is this needed? I would expect the rewrite should only be needed in the math vocabulary.

dmuino: It's needed for an expression like: `foo,1,:eq,(,a,),:cg`

I'm assuming here that this would be blindly added by tools, so they won't know that the expression is not grouped.

atlas-core/src/main/scala/com/netflix/atlas/core/model/MathVocabulary.scala
```diff
@@ -448,6 +449,47 @@ object MathVocabulary extends Vocabulary {
...
+        true
+    }
+
+    override protected def executor: PartialFunction[List[Any], List[Any]] = {
+      case StringListType(keys) :: (expr: Expr) :: stack =>
+        val newExpr = expr.rewrite {
+          case e: DataExpr.GroupBy                   => e.copy(keys = e.keys ::: keys)
+          case nr: NamedRewrite                      => nr.groupBy(keys)
+          case af: AggregateFunction                 => DataExpr.GroupBy(af, keys)
+          case af: AggrMathExpr if af.expr.isGrouped => MathExpr.GroupBy(af, keys)
```
brharrington, line 474: I think this would need to recursively apply it to the underlying expression as well. For example:

```
name,foo,:eq,:sum,(,a,b,),:by,
:max,(,b,),:by,
(,c,),:cg
```

As written, this would turn into:

```
name,foo,:eq,:sum,(,a,b,),:by,
:max,(,b,c,),:by,
```

Rather than:

```
name,foo,:eq,:sum,(,a,b,c,),:by,
:max,(,b,c,),:by,
```

brharrington (changes requested):

atlas-core/src/main/scala/com/netflix/atlas/core/model/DataVocabulary.scala
brharrington: I think the problem is how you are matching the `executor`, if you use `TimeSeriesType(t)`, then it should implicitly convert the query to a data expr using sum as the aggregate.

brharrington (approved):
I pushed a few minor tweaks and some additional tests, seems to be working for the additional cases I could think of

Issue #1355, opened 2021-10-07
Support explicit configurable limits for streaming expressions


brharrington:
Update the eval library so it can be configured with explicit limits for expressions. If an expression exceeds the limit, then have emit an error for that expression and stop evaluating it.

/cc @ruchirj

PR #1397, opened 2021-12-29
1. update eval library so it can be configured with explicit limits for expressions

lavanyachennupati:
### Changes in this pr:
1. update eval library so it can be configured with explicit limits for expressions with config parameter - `atlas.eval.stream.expression-limit`
2. if any expression within the streamContext exceeds the limit, then emit an error to all the datasources using this expression.
3. stop adding additional `AggrDataPoint` to the Aggregator corresponding to an expression when the number of raw data points for that expression coming in exceed the expression limit.
4. drop all `AggrDataPoint` corresponding to the expression within the `TimeGroup` before flushing if raw data points exceeds the configured limit. This is to avoid any user confusion that could be caused by keeping the data points until the limit is exceeded that results in incorrect data.
5. add tests

### Context:
#1355

### Assumptions made this in this pr:
1. `atlas.eval.stream.expression-limit` config parameter applies to all expressions. Could pattern match to apply the limit only to some expressions if we need to support that use case later.
2. `TimeGrouped` seemed like the best place to check if an expression has data points that exceed the configured limit. As TimeGrouped maintains time buffers for every expression, the approach taken here is to stop adding any data points to the Aggregator for an expression when the raw incoming data points exceed the configured limit for a specific timeStamp(floor of the step size of the stream context).
     a) Considering the number of `raw data points` instead of the number of data points after the aggregation for an expression within the buffer. This is to stop adding to the Aggregator earlier if raw data points exceed the limit which I think results in better memory utilization of the buffers. If we were to instead use the number of data points after the aggregation, all data points could still be added(depending on the aggregation function and data) to the expression for a given timestamp until the buffer for that timestamp is to be flushed. Happy to change to use the number of data points after the aggregation if that's desired.
     b) drop all `AggrDataPoint` corresponding to the expression within the `TimeGroup` before flushing for FinalEvaluation, if raw data points exceeds the configured limit. This is to avoid any user confusion that could be caused by keeping the data points until the limit is exceeded that results in incorrect data. Also based on the implementation in this pr(considering the number of raw data points), we might be missing adding any values that come in after the limit is exceeded and might already be present in the aggregation results. 
   c) Only the `AggrDataPoint`s in the buffer for the expression corresponding to the timestamp that exceed the limit are dropped. Any other timestamps in the buffer for this expression might still have data if the number of raw data points are < the limit. This is to avoid completely stopping any future evaluation of the expression because there was an intermittent spike in the data points in the past.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ target
 *.iml
 *.ipr
 *.iws
+project
 
 # anything in .idea dirs
 .idea
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@ target
 *.iml
 *.ipr
 *.iws
-project
 
 # anything in .idea dirs
 .idea
--- a/atlas-eval/src/main/resources/reference.conf
+++ b/atlas-eval/src/main/resources/reference.conf
@@ -15,9 +15,6 @@ atlas.eval {
     // is too old
     num-buffers = 1
 
-    // Limit configured for expressions. Evaluation will be stopped if limit exceeds.
-    expression-limit = 40
-
     // Broad tag keys that should be ignored for the purposes of dropping expensive queries
     // that will be prohibitive to match. These are typically tags that will be applied to
     // everything within a given deployment scope.
--- a/atlas-eval/src/main/resources/reference.conf
+++ b/atlas-eval/src/main/resources/reference.conf
@@ -15,6 +15,9 @@ atlas.eval {
     // is too old
     num-buffers = 1
 
+    // Limit configured for expressions. Evaluation will be stopped if limit exceeds.
+    expression-limit = 40
+
     // Broad tag keys that should be ignored for the purposes of dropping expensive queries
     // that will be prohibitive to match. These are typically tags that will be applied to
     // everything within a given deployment scope.
--- a/atlas-eval/src/main/resources/reference.conf
+++ b/atlas-eval/src/main/resources/reference.conf
@@ -15,6 +15,13 @@ atlas.eval {
     // is too old
     num-buffers = 1
 
+    limits {
+      // Maximum number of raw input datapoints for a data expr. Defaults to Integer.MaxValue
+      max-input-datapoints = 2147483647
+      // Maximum number of datapoints resulting from a group by. Defaults to Integer.MaxValue
+      max-intermediate-datapoints = 2147483647
+    }
+
     // Broad tag keys that should be ignored for the purposes of dropping expensive queries
     // that will be prohibitive to match. These are typically tags that will be applied to
     // everything within a given deployment scope.
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/StreamContext.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/StreamContext.scala
@@ -78,8 +78,12 @@ private[stream] class StreamContext(
 
   def numBuffers: Int = config.getInt("num-buffers")
 
-  // Limit configured for expressions. Evaluation will be stopped if limit exceeds.
-  def expressionLimit: Int = Try(config.getInt("expression-limit")).getOrElse(Integer.MAX_VALUE)
+  // Maximum number of raw input data points for a data expr.
+  def maxInputDatapointsPerExpression: Int = config.getInt("limits.max-input-datapoints")
+
+  // Maximum number of datapoints resulting from a group by for a data expr.
+  def maxIntermediateDatapointsPerExpression: Int =
+    config.getInt("limits.max-intermediate-datapoints")
 
   val interpreter = new ExprInterpreter(rootConfig)
 
@@ -173,6 +177,20 @@ private[stream] class StreamContext(
       val msg = s"rejected expensive query [$query], narrow the scope to a specific app or name"
       throw new IllegalArgumentException(msg)
     }
+  }
+
+  /**
+    * Emit an error to the sources where the number of input
+    * or intermediate datapoints exceed for an expression.
+    */
+  def logDatapointsExceeded(timestamp: Long, dataExpr: DataExpr) = {
+    val diagnosticMessage = DiagnosticMessage.error(
+      s"expression: $dataExpr exceeded the configured max input datapoints limit" +
+      s" '$maxInputDatapointsPerExpression' or max intermediate" +
+      s" datapoints limit '$maxIntermediateDatapointsPerExpression'" +
+      s" for timestamp '$timestamp}"
+    )
+    log(dataExpr, diagnosticMessage)
   }
 
   /**
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/StreamContext.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/StreamContext.scala
@@ -78,6 +78,9 @@ private[stream] class StreamContext(
 
   def numBuffers: Int = config.getInt("num-buffers")
 
+  // Limit configured for expressions. Evaluation will be stopped if limit exceeds.
+  def expressionLimit: Int = Try(config.getInt("expression-limit")).getOrElse(Integer.MAX_VALUE)
+
   val interpreter = new ExprInterpreter(rootConfig)
 
   def findBackendForUri(uri: Uri): Backend = {
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
@@ -44,7 +44,7 @@ private[stream] class TimeGrouped(
   max: Int
 ) extends GraphStage[FlowShape[AggrDatapoint, TimeGroup]] {
 
-  type AggrMap = scala.collection.mutable.AnyRefMap[DataExpr, AggrDatapoint.Aggregator]
+  type AggrMap = scala.collection.mutable.AnyRefMap[DataExpr, (AggrDatapoint.Aggregator, Boolean)]
 
   /**
     * Number of time buffers to maintain. The buffers are stored in a rolling array
@@ -62,6 +62,10 @@ private[stream] class TimeGrouped(
   private val metricName = "atlas.eval.datapoints"
   private val droppedOld = context.registry.counter(metricName, "id", "dropped-old")
   private val droppedFuture = context.registry.counter(metricName, "id", "dropped-future")
+
+  private val droppedLimitExceeded =
+    context.registry.counter(metricName, "id", "dropped-limit-exceeded")
+
   private val buffered = context.registry.counter(metricName, "id", "buffered")
 
   private val clock = context.registry.clock()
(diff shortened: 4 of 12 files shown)


atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/TimeGroupedSuite.scala
```diff
@@ -336,4 +346,29 @@ class TimeGroupedSuite extends FunSuite {
     val groups = run(data)
     assertEquals(groups, List(TimeGroup(10, step, expected)))
   }
+
+  test("group by aggregate with additional tags: count") {
```
lavanyachennupati, line 350: Added this test to understand the desired behavior. 

i.e, if the `DataExpr` is `DataExpr.GroupBy(DataExpr.Count(Query.True), List("category"))`, wondering if data points with additional tags like `blah` other than `category` would be rolled up or not when aggregating.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
```diff
@@ -97,8 +99,18 @@ private[stream] class TimeGrouped(
...
         if (!v.isHeartbeat) {
           buf(i).get(v.expr) match {
-            case Some(aggr) => aggr.aggregate(v)
-            case None       => buf(i).put(v.expr, AggrDatapoint.newAggregator(v))
+            case Some(aggr) => {
+              // drop the data points if an expression exceeds the configured limit within the time buffer and stop any final evaluation of this expression.
+              // emit an error to all data sources that use this particular data expression.
+              if (aggr.numRawDatapoints > expressionLimit) {
+                // emit an error to the source
+                val diagnosticMessage = DiagnosticMessage.error(
```
manolama, line 107: Is there a way to null out or invalidate the aggregator when the condition is met? That way it can log once and avoid subsequent object generation and logging?

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
```diff
@@ -115,10 +127,14 @@ private[stream] class TimeGrouped(
       }
 
       private def toTimeGroup(ts: Long, aggrMap: AggrMap): TimeGroup = {
+        val aggregateMapWithExpressionLimits =
+          aggrMap.filter(t => t._2.numRawDatapoints < expressionLimit).toMap
```
manolama, line 131: If sliding window functions use this data, I wonder if it would best to avoid emitting a value for such functions any time a bucket has exceeded the limit instead of returning a value that is missing some data.

atlas-eval/src/main/resources/reference.conf
```diff
@@ -15,6 +15,9 @@ atlas.eval {
     // is too old
     num-buffers = 1
 
+    // Limit configured for expressions. Evaluation will be stopped if limit exceeds.
+    expression-limit = 40
```
manolama, line 19: Should this be `50k` like the UT below?

manolama (commented):
Is the purpose of the limit to reduce cardinality for a given expression and void creating big data structures? Or is it just a limit on the number of raw values on a given expression?

atlas-eval/src/main/resources/reference.conf
lavanyachennupati: was changing the limits to run the tests with different values and missed to set it to a high value. 50k sounds like a reasonable starting point to me. Curious to know if there's a better suited limit based on load testing or past experience.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
lavanyachennupati: Not sure I completely understand.

Say for simplicity if there's only one expression being evaluated for the datasource, the behavior in this pr would return `TimeGroup(t, step, Map.empty)`, for the timestamp `t` if the limit is exceeded for that expression. And the `FinalExprEval` will not evaluate this `TimeGroup` for the expression and will result in a `TimeSeriesMessage` with `NO DATA` to all the datasources of this expression.

These test cases might be a bit clear in explaining the above behavior.
1. https://github.com/Netflix/atlas/pull/1397/files#diff-35ea10e16e6ff84c8039f99941538934bd9e04ff1464befdf0674c6e441389fbR158
2. https://github.com/Netflix/atlas/pull/1397/files#diff-c63398ad74c3a6c5e7f0c9baa4df53ae37e323038fe9ba5853383b5d6b615484R127

Would there be anything specific to sliding functions using this data that might not be the desired behavior?

manolama: I didn't dig into the `FinalExprEval` but my thinking was that if someone adds an `rolling-sum` function that maybe just the limited `TimeGroups` would be skipped and others included. If _any_ time groups are limited and the final expression returns the `NO DATA` result that should be OK.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
```diff
@@ -62,6 +62,10 @@ private[stream] class TimeGrouped(
   private val metricName = "atlas.eval.datapoints"
   private val droppedOld = context.registry.counter(metricName, "id", "dropped-old")
   private val droppedFuture = context.registry.counter(metricName, "id", "dropped-future")
+
+  private val droppedLimitExceeded =
+    context.registry.counter(metricName, "id", "dropped-limit-exceeded")
```
lavanyachennupati, line 67: this wouldn't give insights into the rate of messages dropped because of limit being exceeded  broken down by the expression (if there are multiple expressions being evaluated). The same I believe is relevant to `dropped-old` and `dropped-future`. Adding expression as a tag I think would blow up the cardinality.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
```diff
@@ -99,18 +103,21 @@ private[stream] class TimeGrouped(
...
-              if (aggr.numRawDatapoints > expressionLimit) {
+              if (!limitExceeded && aggr.numRawDatapoints >= expressionLimit) {
                 // emit an error to the source
                 val diagnosticMessage = DiagnosticMessage.error(
-                  s"expression exceeded the configured limit '$expressionLimit' for timestamp '${timestamps(i)}"
+                  s"expression: ${v.expr} exceeded the configured limit '$expressionLimit' for timestamp '${timestamps(i)}"
                 )
                 context.log(v.expr, diagnosticMessage)
-              } else aggr.aggregate(v)
+                buf(i).put(v.expr, (aggr, true))
```
lavanyachennupati, line 115: I think there's potential to optimize heap usage by deleting all the existing `AggrDataPoint`s in the `Aggregator` once the limit for the expression has exceeded instead of having the limit number of `AggrDataPoint`s be GC'd eventually. While I'm trying to understand the use cases/problems this [issue](https://github.com/Netflix/atlas/issues/1355) is trying to solve, I haven't made any changes to introduce a `delete(datapoint: AggrDataPoint)` on the `Aggregator` which probably is not needed.

.gitignore
```diff
@@ -21,6 +21,7 @@ target
 *.iml
 *.ipr
 *.iws
+project
```
brharrington, line 24: Why was this added? There are build files there that are occasionally updated, so I don't think that directory should be ignored.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/StreamContext.scala
```diff
@@ -78,6 +78,9 @@ private[stream] class StreamContext(
 
   def numBuffers: Int = config.getInt("num-buffers")
 
+  // Limit configured for expressions. Evaluation will be stopped if limit exceeds.
+  def expressionLimit: Int = Try(config.getInt("expression-limit")).getOrElse(Integer.MAX_VALUE)
```
brharrington, line 82: A try shouldn't be needed here, the `reference.conf` should just default to the max value.

https://github.com/lightbend/config#how-to-handle-defaults

The name `expression-limit` makes me think it is limiting the number of expressions rather than the amount of data for an expression. In theory we could have additional limits in the future. Maybe something like:

```
limits {
  // Maximum number of raw input datapoints for a data expr
  max-input-datapoints = 100k

  // Maximum number of datapoints resulting from a group by 
  max-intermediate-datapoints = 10k
}
```

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
brharrington: Agreed, the metric should not track per expression. There should be a message emitted to the user to indicate there is a problem with a given expression.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
```diff
@@ -97,8 +103,21 @@ private[stream] class TimeGrouped(
       private def aggregate(i: Int, v: AggrDatapoint): Unit = {
         if (!v.isHeartbeat) {
           buf(i).get(v.expr) match {
-            case Some(aggr) => aggr.aggregate(v)
-            case None       => buf(i).put(v.expr, AggrDatapoint.newAggregator(v))
+            case Some((aggr, limitExceeded)) => {
+              // drop the data points if an expression exceeds the configured limit within the time buffer and stop any final evaluation of this expression.
```
brharrington, line 107: wrap to 100 chars

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
```diff
@@ -43,14 +44,15 @@ private[stream] class TimeGrouped(
   max: Int
 ) extends GraphStage[FlowShape[AggrDatapoint, TimeGroup]] {
 
-  type AggrMap = scala.collection.mutable.AnyRefMap[DataExpr, AggrDatapoint.Aggregator]
+  type AggrMap = scala.collection.mutable.AnyRefMap[DataExpr, (AggrDatapoint.Aggregator, Boolean)]
```
brharrington, line 47: It might be cleaner to just have the aggregator check and enforce the limits.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
brharrington: We should probably revisit after figuring out whether or not to disable the expression once the limit is hit. If it is disabled and removed from being checked, then it should handle the cleanup, but we would need to figure out exactly how that would work.

atlas-eval/src/test/scala/com/netflix/atlas/eval/stream/TimeGroupedSuite.scala
brharrington: In a real scenario the datapoint will only have tags that either have an exact match in the query clause or are listed in the group by. So the set below would be invalid since there are no `:eq` clauses in the query and `blah` is not in the group by list.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/StreamContext.scala
lavanyachennupati: Thanks for pointing to the docs and the reference example!

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/TimeGrouped.scala
lavanyachennupati: I took a stab at making these changes in the recent commit. And I think it looks much cleaner and handles other edge cases that I missed in the earlier commits. 

One such case I missed: `FinalEvalExpr` aggregates all `AggrDataPoint`s in a `TimeGroup` again https://github.com/Netflix/atlas/blob/master/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/FinalExprEval.scala#L173.

brharrington (approved):

PR #1401, opened 2022-01-05
Add encode methods to `PublishPayloads` for `List[Datapoint]`s.

manolama:
These are used for proxies and operations that do not require computing
a hash on the data point identity.

### DIFF
--- a/atlas-webapi/src/main/scala/com/netflix/atlas/webapi/PublishPayloads.scala
+++ b/atlas-webapi/src/main/scala/com/netflix/atlas/webapi/PublishPayloads.scala
@@ -355,6 +355,14 @@ object PublishPayloads {
     }
   }
 
+  private def encodeBatchDatapoint(gen: JsonGenerator, d: Datapoint): Unit = {
+    gen.writeStartObject()
+    encodeTags(gen, d.tags)
+    gen.writeNumberField("timestamp", d.timestamp)
+    gen.writeNumberField("value", d.value)
+    gen.writeEndObject()
+  }
+
   /**
     * Encode batch of datapoints.
     */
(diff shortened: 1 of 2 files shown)


atlas-webapi/src/test/scala/com/netflix/atlas/webapi/PublishPayloadsSuite.scala
```diff
@@ -27,7 +27,7 @@ class PublishPayloadsSuite extends FunSuite {
 
   private val timestamp = 1636116180000L
 
-  private def datapoints(n: Int): List[DatapointTuple] = {
+  private def datapointTuples(n: Int): List[DatapointTuple] = {
```
brharrington, line 30: To reduce code duplication, can you update this to be something like:

```scala
private def datapointTuples(n: Int): List[DatapointTuple] = {
  datapoints(n).map(_.toTuple)
}
```

atlas-webapi/src/main/scala/com/netflix/atlas/webapi/PublishPayloads.scala
```diff
@@ -355,6 +355,22 @@ object PublishPayloads {
     }
   }
 
+  def encodeBatchDatapoint(gen: JsonGenerator, d: Datapoint): Unit = {
```
brharrington, line 358: It seems like this could be made private.

atlas-webapi/src/main/scala/com/netflix/atlas/webapi/PublishPayloads.scala
```diff
@@ -355,6 +355,22 @@ object PublishPayloads {
...
+    gen.writeStartObject()
+    encodeTags(gen, d.tags)
+    gen.writeNumberField("timestamp", d.timestamp)
+    gen.writeNumberField("value", d.value)
+    gen.writeEndObject()
+  }
+
+  def encodeBatchDatapoint(d: Datapoint): String = {
```
brharrington, line 366: It doesn't look like this is used.

brharrington (approved):
A few minor suggestions, but mostly looks good to me.

atlas-webapi/src/main/scala/com/netflix/atlas/webapi/PublishPayloads.scala
manolama: Just going for parity with the tuple APIs here.

PR #1402, opened 2022-01-05
limit visibility of methods for single datapoint

brharrington:
Update PublishPayloads helper to limit the visibility
for methods related to a single datapoint. These are
only needed internally or in some cases used as part
of the unit tests.


manolama (approved):
LGTM

PR #1403, opened 2022-01-06
add container and process to allowed keys

brharrington:
Support for the Titus Multi-Container Tagging
Specification.


copperlight (approved):
Looks good - covers the nf.container and nf.process tags.

PR #1404, opened 2022-01-07
track number of streams dropping data

brharrington:
Updates the stream metadata to keep track of the number
of recently received and dropped messages rather than
the number overall. Currently this is done based on a
step interval. Adds a guage to show the number of streams
with a dimension indicating the state.


manolama (approved):
Looks straightforward to me. Sounds like the `StepLong` is a windowed counter.

PR #1405, opened 2022-01-07
make JsonSupport.toJson final

brharrington:
Users should override `encode` to customize and can
get confusing results if only `toJson` is overridden.
Also updates the `toJson` to ensure the generator will
be cleaned up if `encode` throws an exception.


manolama (approved):
LGTM

PR #1406, opened 2022-01-22
Fix a bug regarding chained group by queries where an error was thrown

manolama:
if a subsequent key group was not a `slice` of the previous group but
was still a subset of that group. E.g.
`(,name,mode,value,),:by,(,name,value,),:by` would throw the error
despite `name` and `value` being present in the original group.

### DIFF
--- a/atlas-core/src/test/scala/com/netflix/atlas/core/model/MathGroupBySuite.scala
+++ b/atlas-core/src/test/scala/com/netflix/atlas/core/model/MathGroupBySuite.scala
@@ -87,6 +87,23 @@ class MathGroupBySuite extends FunSuite {
     assertEquals(rs, expected)
   }
 
+  test("(,name,mode,value,),:by,(,name,value,),:by") {
+    val input = List(
+      ts(1),
+      ts(2),
+      ts(3)
+    )
+    val rs = groupBy(input, List("name", "mode", "value"), List("name", "value"), MathExpr.Sum)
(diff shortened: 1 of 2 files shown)


dmuino (approved):

atlas-core/src/test/scala/com/netflix/atlas/core/model/MathGroupBySuite.scala
```diff
@@ -87,6 +87,24 @@ class MathGroupBySuite extends FunSuite {
...
 
+  test("(,name,mode,value,),:by,(,name,value,),:by") {
+    val input = List(
+      ts(1),
+      ts(2),
+      ts(3)
+    )
+    val rs = groupBy(input, List("name", "mode", "value"), List("name", "value"), MathExpr.Sum)
+    assertEquals(rs.size, 3)
+    System.out.println(rs)
```
brharrington, line 98: Please remove the println to reduce noise when running tests. Also, as an FYI, in scala you can just do `println(rs)`.

brharrington:
Looks like you need to run `make format` to clear up the formatting issues.

atlas-core/src/test/scala/com/netflix/atlas/core/model/MathGroupBySuite.scala
manolama: doh

brharrington (approved):

PR #1412, opened 2022-02-08
fix performance of limit check

brharrington:
For group by it was calling `datapoints.size` which is a
linear operation and maps the values resulting in a lot
of allocations. This changes the intermediate size check
to be a constant time operation.

### DIFF
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
@@ -92,7 +92,7 @@ object AggrDatapoint {
     * instance have the same data expression.
     */
   trait Aggregator {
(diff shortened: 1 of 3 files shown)


atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
```diff
@@ -174,20 +179,25 @@ object AggrDatapoint {
...
+          inputDatapointCounter += 1
           aggregator
         case _ =>
           throw new IllegalArgumentException("datapoint is not for a grouped expression")
       }
     }
 
+    // This should be a constant time operation as it will be used to check the limit
+    // for incoming datapoints.
+    override def numIntermediateDatapoints: Int = aggregators.size
```
lavanyachennupati, line 191: Ah nice catch! Thank you!

lavanyachennupati (approved):

PR #1414, opened 2022-02-09
reduce allocations for datapoint aggregators

brharrington:
Before the counter and associated id was being created
each time an aggregator instance was. Now it is part of
a settings object which can be reused. The aggregators
have also been refactored a bit to reduce code duplication.


lavanyachennupati (approved):
Much cleaner 👍

PR #1415, opened 2022-02-11
fix memory leak for failed streams

brharrington:
The BroadcastHub doesn't complete if there are still items
pending. When the subscriber failed, the failure wasn't
propagated up since the BroadcastHub can have another
subscriber come along later. Change it to use a publisher
with no fanout, this will fail along with the connected
downstream.


manolama (approved):
Good catch, LGTM

PR #1423, opened 2022-03-30
lwcapi: list streams if path ends with `/`

brharrington:
Update the matching logic to list for the end of the path
or a single slash. Before it would try to use an empty string
as the id which can be confusing.


manolama (approved):
Thanks!

PR #1428, opened 2022-04-09
drop use of scala reflection for atlas-core

brharrington:
Scala reflection libraries were removed in scala 3.
The only usage was for loading rule instances based on
a config entry. This can be done easily enough with the
normal java reflection.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/validation/Rule.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/validation/Rule.scala
@@ -20,9 +20,6 @@ import com.netflix.atlas.core.util.SortedTagMap
 import com.netflix.spectator.api.Id
 import com.typesafe.config.Config
 
-import scala.reflect.runtime.universe
-import scala.util.control.NonFatal
-
 /**
   * Base type for validation rules.
   */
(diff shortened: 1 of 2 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/validation/Rule.scala
```diff
@@ -72,39 +69,8 @@ object Rule {
...
   def load(ruleConfigs: List[_ <: Config], useComposite: Boolean): List[Rule] = {
-    val configClass = classOf[Config]
     val rules = ruleConfigs.map { cfg =>
-      val cls = Class.forName(cfg.getString("class"))
-
-      try {
-        cls.getConstructor(configClass).newInstance(cfg).asInstanceOf[Rule]
-      } catch {
-        case NonFatal(th) =>
-          val runtimeMirror = universe.runtimeMirror(cls.getClassLoader)
```
manolama, line 83: +1 for getting rid of this. Took 30s on a host I was playing with before joining.

manolama (approved):

atlas-core/src/main/scala/com/netflix/atlas/core/validation/Rule.scala
brharrington: What took 30s? I haven't seen any problems with that particular code, but have never really understood universe/mirror stuff all that well.

manolama: Copying and walking the mirror when running in standalone mode. Actually it may not have been that it took a long time just that it created a ton of garbage on startup and it ran fine after the first collection.

PR #1429, opened 2022-04-09
fix compile and test errors on scala3

brharrington:
Fixes some minor issues that cause compile errors on
scala3 for atlas-core. The atlas-core project now
compiles and all tests pass except for AlgoStateSuite
due to issues with atlas-json.


manolama (approved):
Pretty nice that this is all that's required for Atlas to build under 3.

PR #1431, opened 2022-04-13
lwc: use asg for matching exprs to a cluster

brharrington:
Some expressions have an exact match for `nf.asg`, but do
not have conditions for `nf.app` or `nf.cluster`. This change
will parse the asg value based on the server group naming
rules and add a cluster restriction so that expressions for
a specific asg will only show when requesting data for the
cluster containing that asg.

### DIFF
--- a/atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionSplitter.scala
+++ b/atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionSplitter.scala
@@ -189,9 +189,9 @@ class ExpressionSplitter(config: Config) {
         // Rewrite exact match for nf.asg to an exact match for the nf.cluster. When
         // fetching expressions for a cluster, this will allow expressions with only
         // a check for the asg to get filtered appropriately.
-        case Query.Equal("nf.asg", asg) =>
+        case q @ Query.Equal("nf.asg", asg) =>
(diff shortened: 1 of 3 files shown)


atlas-lwcapi/src/main/scala/com/netflix/atlas/lwcapi/ExpressionSplitter.scala
```diff
@@ -184,8 +184,20 @@ class ExpressionSplitter(config: Config) {
...
-    val tmp = expr.rewrite { case kq: KeyQuery if !keepKeys.contains(kq.k) => Query.True }
-    simplify(tmp.asInstanceOf[Query])
+    val query = expr
+      .rewrite {
+        // Rewrite exact match for nf.asg to an exact match for the nf.cluster. When
+        // fetching expressions for a cluster, this will allow expressions with only
+        // a check for the asg to get filtered appropriately.
+        case Query.Equal("nf.asg", asg) =>
+          val cluster = ServerGroup.parse(asg).cluster()
+          Query.Equal("nf.cluster", cluster)
```
manolama, line 194: Is it worth checking for a malformed ASG that could cause the cluster to return null? What would happen downstream in the evaluator if `cluster` is `null`?

manolama (approved):
Good to go if the `null` `cluster` possibility is ok. I could see something mangling the ASG on accident.

PR #1434, opened 2022-04-15
allow mixed grouping for custom averages

brharrington:
In the past this wasn't permitted because binary operations
had to have the same grouping on both sides. In #1244 that
was changed so that one side could have a subset of the other.


manolama (approved):
Just have to watch out for massive cross-product joins. Should double check that there is a limiter that would catch it if the num and denom snuck under the time series limit.

brharrington:
There isn't a massive cross-product here, the matching of the sides is based on matching keys, so it is an inner join.

PR #1435, opened 2022-04-15
lwc: update operator validation

brharrington:
Exclude operators that are confusing when used in a
streaming context. Specifically:

- `integral`: accumulates the value over the life of
  the evaluation. For a streaming evaluation the user
  has no way to know the time window it has been
  accumulating the value over.
- _Filters_: all filtering operators rely on statistics
  from the evaluation context to select the results.
  For streaming evaluation, that context is always a
  single step interval. Filtering is best left to the
  consumer of the data.


manolama (approved):
LGTM. Will the alert UI call out to a validation API before a user save's their config to see if it would throw the IAE?

brharrington:
Yes, they call validate.

PR #1443, opened 2022-05-05
GitHub Actions ubuntu-latest

sullis:


brharrington:
What is the rationale for this change? MacOS is used intentionally because some of the tests render graph images for comparison and there can be subtle differences for things like font rendering between systems..

sullis:
> What is the rationale for this change? MacOS is used intentionally because some of the tests render graph images for comparison and there can be subtle differences for things like font rendering between systems..

Interesting! I wasn't aware of that.  I'll close this.

PR #1455, opened 2022-08-28
switch to using Spring for DI

brharrington:
Since Netflix is moving from Guice to Spring, switch the
DI framework to use Spring. This is mostly just migrating
the Guice modules.


atlas-spring-akka/src/main/scala/com/netflix/atlas/akka/AkkaConfiguration.scala
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+  * Configures the actor system and web server. This module expects that bindings
+  * are available for `com.typesafe.config.Config` and `com.netflix.spectator.api.Registry`.
+  */
+@Configuration
+class AkkaConfiguration {
+
+  @Bean
+  def actorySystemService(config: Optional[Config]): ActorSystemService = {
```
copperlight, line 38: `actorySystemService`?

PR #1465, opened 2022-09-09
improve couldMatch check for not clauses

brharrington:
If the inverted query has an exact match, then adding dimensions cannot change the result and it should be excluded.


manolama:
LGTM

PR #1470, opened 2022-09-27
Add a `:contains` predicate filter for compatibility with internal lo…

manolama:
…g queries.

Not meant to be used for metrics as it executes a full scan. Bump Spectator to 1.3.8

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/model/QueryVocabulary.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/model/QueryVocabulary.scala
@@ -19,6 +19,7 @@ import com.netflix.atlas.core.stacklang.SimpleWord
 import com.netflix.atlas.core.stacklang.StandardVocabulary
 import com.netflix.atlas.core.stacklang.Vocabulary
 import com.netflix.atlas.core.stacklang.Word
+import com.netflix.spectator.impl.matcher.PatternUtils
 
 object QueryVocabulary extends Vocabulary {
 
@@ -39,6 +40,7 @@ object QueryVocabulary extends Vocabulary {
     GreaterThanEqual,
     Regex,
     RegexIgnoreCase,
+    Contains,
     In,
     And,
     Or,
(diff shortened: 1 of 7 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/model/QueryVocabulary.scala
```diff
@@ -329,6 +330,38 @@ object QueryVocabulary extends Vocabulary {
       List("name,DiscoveryStatus_(UP|DOWN)", "name,discoverystatus_(Up|Down)", "ERROR:name")
   }
 
+  case object Contains extends KeyValueWord {
+
+    override def name: String = "contains"
+
+    def newInstance(k: String, v: String): Query = Query.Contains(k, v)
```
brharrington, line 337: Do we need to keep this as a separate query type and plumb it through everywhere? I'm wondering if it would be cleaner to just add the word and then create an equivalent regex pattern such as:

```
def newInstance(k: String, v: String): Query = Query.Regex(k, s".*${PatternUtils.escape(v)}")
```

atlas-core/src/main/scala/com/netflix/atlas/core/model/Query.scala
```diff
@@ -396,6 +398,17 @@ object Query {
     override def toString: String = s"$k,$v,:reic"
   }
 
+  case class Contains(k: String, v: String) extends PatternQuery {
+    override def pattern: PatternMatcher = PatternMatcher.compile(PatternUtils.escape(v))
+
+    def check(s: String): Boolean = pattern.matches(s)
```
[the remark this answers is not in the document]
manolama: ack

atlas-core/src/main/scala/com/netflix/atlas/core/model/QueryVocabulary.scala
manolama: Ack. We can try it this way for now. We'll just have to remember that re-serializing the query will return the regex instead.

brharrington (approved):

PR #1474, opened 2022-10-25
update normalize behavior for UI

brharrington:
Updates the behavior of normalize to reflect preferences of the UI team. In particular:

- It will not try to extract common query aspects into a `:cq` clause.
- The helpers for filtering based on summary stats will be preserved rather than expanding.

/cc @nathfisher


manolama (approved):
LGTM if no one else is using it for anything important.

brharrington:
Even if there is usage, it shouldn't break anything. We don't guarantee the exact updates, just that it should be semantically equivalent to the input and that semantically equivalent queries should typically result in the same normalized version.

PR #1488, opened 2022-11-12
lwc: use SortedTagMap for data points

brharrington:
The main advantage is that for eval it can just quickly copy into an array without all the hashing and extra space in the hash array to avoid collisions.

Note, this change is safe to roll out as long as the lwcapi service is pushed first.


manolama (approved):
Should be fine assuming we update everything that generates LwcMessages first.

Issue #1047, opened 2019-06-05
Y-Axis labels as units-of-time rather than SI units.


osi:
When graphing timers, the Y axis is in seconds, represented as SI units.

To be more human-friendly, the Y axis should use SI units for values less than a second, and use minutes/hours/days for values greater than a second.

PR #1497, opened 2022-11-22
Grapher: Add support for duration ticks with a base of seconds. Fixes…

manolama:
… #1047

by adding support for `&tick_labels=duration` that will scale ticks from picoseconds to years.
Also change the legend behavior so that counts are no longer tagged with the same prefix as the other stats to differentiate in binary and duration mode.


atlas-chart/src/main/scala/com/netflix/atlas/chart/model/TickLabelMode.java
```diff
@@ -40,7 +40,12 @@ public enum TickLabelMode {
    *
    * https://en.wikipedia.org/wiki/Binary_prefix
    */
-  BINARY;
+  BINARY,
+
+  /**
+   * Use SI for durations less than a second and 's', 'min', 'hr', 'day', 'mon', 'yr' over that.
```
brharrington, line 46: We should try to keep these consistent with the indicators for [time parameters](https://netflix.github.io/atlas-docs/api/time-parameters/#simple-duration) where possible.

Do we need months? I think seconds, minutes, hours, days, weeks, and years should suffice. The main reason is that since these are durations I think it is clear to most users the quantity of time it represents where a month is ambiguous in that regard though in other places we treat month as "30 days". Weeks, `w`, also has the benefit of being less overloaded for the unit indicator than `m/M` is (milli, mega, minute, month, etc). Years technically is also ambiguous in this way, but is less varied and less likely to come up in a lot of usage.

Also, for large values, say 10k+ years, would it use exponential representation? It seems like it could be:

- Sub-second, metric prefix + `s`, for example: `ns` or `ms`. Matches current behavior in this PR. 
- The `s`, `h`, `w` (see below) aren't used elsewhere so work for larger values with no ambiguity.
- For year, `y` could be confused with yocto though that prefix doesn't come up much. The use of `yr` is probably fine to avoid that even though it doesn't match the time parameter.
- The time parameters use `m` for minute, but that could be confused for `milli`. There are only 3600 seconds in a minute, so we could just skip minutes and go from seconds to hours. I think the main argument for it is that it may be easier for a user to reason about 15m vs 900s.
- If something is too big to fit in years, then I think it is fine to just switch to an exponential axis the way we do with binary labels. These are pretty rare in practice.

atlas-chart/src/test/scala/com/netflix/atlas/chart/graphics/TicksSuite.scala
```diff
@@ -400,6 +400,239 @@ class TicksSuite extends FunSuite {
...
+  }
+
+  test("duration [1.0e-12, 1.0e12]") {
+    val ticks = Ticks.duration(1.0e-12, 1.0e12, 5)
+    sanityCheck(ticks)
+    assertEquals(ticks.size, 20)
+    assertEquals(ticks.count(_.major), 5)
+    assertEquals(ticks.head.offset, 0.0)
+    assertEquals(ticks.head.label, "1585.490yr")
+    assertEquals(ticks.last.label, "31709.792yr")
```
brharrington, line 623: This looks like it would be too many characters to fit in the space allocated for a tick label. I also wonder if for the duration access it should act more like the binary mode and not have a decimal place. The tick marks would be on whole numbers for the given unit.

atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/Ticks.scala
```diff
@@ -115,6 +115,49 @@ object Ticks {
...
+          (major * f, minor * f, minorPerMajor)
+      }
+    }
+
+    val majorMultiples = List(
+      List(1, 2, 3, 4, 5, 6, 10, 15, 30, 60),
+      List(1, 2, 3, 4, 5, 6, 10, 15, 30, 3600),
+      List(1, 2, 3, 4, 6, 12, 3600 * 24),
+      List(1, 2, 5, 10, 15, 86400 * 30),
+      List(1, 2, 3, 4, 6, 86400 * 30 * 365)
```
brharrington, line 133: 30 years? Was this supposed to be `86400 * 365`?

atlas-chart/src/main/scala/com/netflix/atlas/chart/model/TickLabelMode.java
manolama: Yeah, we can switch to weeks. Months is a bit more intuitive for those large values but you're right, the 30d fudging would add up to towards the end of the year.

I do want to keep minutes for reasoning. We could probably squeeze in `min` as the prefix for minutes and try to truncate the decimal values a bit.

atlas-chart/src/test/scala/com/netflix/atlas/chart/graphics/TicksSuite.scala
manolama: Ack, I'll try that out.

atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/Ticks.scala
manolama: This is why we have code reviews ;)

brharrington (commented):
Looks pretty good, but I did find some examples that I think could be improved. 

duration

```
/api/v1/graph?q=45,300&tick_labels=duration
```

Y-axis only shows `m`, this could be confused with milli. Avoiding the decimal place for the duration axis would at least provide a subtle indication.

duration

```
/api/v1/graph?q=45,300e5&tick_labels=duration
```

Here is another example where I think avoiding decimal places and aligning ticks to whole weeks would be easier to reason about. That is `17w`, `34w`, etc is probably easier to reason about than `17.381w` and `34.762w`.

duration

```
/api/v1/graph?q=45,300e3&tick_labels=duration
```

Year indication for total throws off the count alignment. It also mixes units on the y-axis. Typically we try to use a single unit for the tick labels. So `0.5d` would be preferred to `12h` on the bottom tick label. If we want to go with whole numbers for the duration axis, I think this example would still work fine with sticking with hours throughout or using days and keeping the major ticks at the day boundary.
 
duration

```
/api/v1/graph?q=45,300e6&tick_labels=duration
```

No tick labels shown with a large range. Also, the alignment for the legend stats on the second legend entry is off from the first entry because of the `yr` prefix. Similar to total on previous example.

duration

```
/api/v1/graph?q=45,300e4&tick_labels=duration
```

Decimal places for the top tick is inconsistent with the others, that could be confused with having a much larger magnitude.

manolama:
Ack, I was thinking more along the lines of Grafana with the change in prefix but I can set it to Atlas style.

atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/LegendEntry.scala
```diff
@@ -80,11 +80,30 @@ case class LegendEntry(styles: Styles, plot: PlotDef, data: DataDef, showStats:
           val avg = format(stats.avg)
           val last = format(stats.last)
           val total = format(stats.total)
-          val count = format(stats.count)
+          val count = format(stats.count, false)
+
           val rows = List(
-            "    Max : %s    Min  : %s".format(max, min),
```
brharrington, line 85: I think you could fix the alignment by just specifying the width and left alignment on the format string. For example:

```
scala> "  Min : %-8s  Max : %-8s".format("1.00a", "1.00b")
val res4: String = "  Min : 1.00a     Max : 1.00b   "

scala> "  Min : %-8s  Max : %-8s".format("1.00ab", "1.00b")
val res5: String = "  Min : 1.00ab    Max : 1.00b   "
```

atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/Ticks.scala
```diff
@@ -174,16 +227,28 @@ object Ticks {
...
 
+  private def getDurationPrefix(v: Double, major: Double): UnitPrefix = {
+    val m = UnitPrefix.durationRange(major)
+    if (v <= m.factor) m else UnitPrefix.durationRange(v)
+  }
+
+  private def durationLabelFormat(prefix: UnitPrefix, v: Double): String = {
+    val f = if (v < 1) v / prefix.factor else v
+    (f * 1000.0).toInt match {
+      case _ if v >= 3.1536e10 => "%.1e%s" // 100+ years switch to exponent
```
brharrington, line 238: `>= 1000 years`?

manolama: doh

atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/LegendEntry.scala
manolama: ack, lemme try it.

brharrington (approved):
Noticed the alignment was still off, pushed a minor fix to increase the width on the format specifier so there would be enough room.

PR #1500, opened 2022-11-29
add value function to dedup inputs

brharrington:
When pushing data to memory the blocks can easily overwrite when an update for a given datapoint is received. In other use-cases this can be quite expensive and undesirable. This change adds a value function that can be used to buffer the updates for a window of time and dedup the data.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/norm/DedupValueFunction.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/norm/DedupValueFunction.scala
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2014-2022 Netflix, Inc.
+ *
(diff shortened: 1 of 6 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/norm/DedupValueFunction.scala
```diff
@@ -0,0 +1,71 @@
...
+    if (timestamp == times(i)) {
+      values(i) = value
+    } else {
+      flushBefore(times(i))
+      times(i) = timestamp
+      values(i) = value
+    }
+  }
+
+  private def flushBefore(cutoff: Long): Unit = {
```
manolama, line 49: nit: may want to rename it `flushTo` since it's inclusive of the cutoff.

manolama (approved):

PR #1509, opened 2023-01-13
Add a unique colors list to the palettes for use by heatmaps.

manolama:
Also add an equals override to palettes for unit testing.

### DIFF
--- a/atlas-chart/src/main/scala/com/netflix/atlas/chart/model/Palette.scala
+++ b/atlas-chart/src/main/scala/com/netflix/atlas/chart/model/Palette.scala
@@ -75,7 +75,7 @@ case class Palette(name: String, colorArray: ArraySeq[Color]) {
     *   A deterministic color in the palette.
     */
   def colors(i: Int): Color = {
-    val index = Math.abs(i - ((i / colorArray.length) * colorArray.length))
+    val index = math.abs(i) % colorArray.length
     colorArray(index)
   }
 }
--- a/atlas-chart/src/main/scala/com/netflix/atlas/chart/model/Palette.scala
+++ b/atlas-chart/src/main/scala/com/netflix/atlas/chart/model/Palette.scala
@@ -18,67 +18,39 @@ package com.netflix.atlas.chart.model
 import java.awt.Color
 import java.io.FileNotFoundException
 import java.util.concurrent.ConcurrentHashMap
-
 import com.netflix.atlas.chart.Colors
 import com.netflix.atlas.core.util.Strings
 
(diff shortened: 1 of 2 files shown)


atlas-chart/src/main/scala/com/netflix/atlas/chart/model/Palette.scala
```diff
@@ -24,6 +24,30 @@ import com.netflix.atlas.core.util.Strings
 
 case class Palette(name: String, colors: Int => Color) {
```
brharrington, line 25: Given that all implementations we have are based on a simple array of colors, I would suggest just changing this to:

```
case class Palette(name: String, colors: ArraySeq[Color])
```

The colors sequence could then be used directly for heatmap or other similar use-cases.

manolama: Ack, I'll take a look

atlas-chart/src/main/scala/com/netflix/atlas/chart/model/Palette.scala
```diff
@@ -64,9 +60,24 @@ case class Palette(name: String, colors: Int => Color) {
...
+    * Rotates through the colors in the palette based on the index, returning a
+    * deterministic color.
+    *
+    * @param i
+    *   A positive integer value.
+    * @return
+    *   A deterministic color in the palette.
+    */
+  def colors(i: Int): Color = {
+    val index = Math.abs(i - ((i / colorArray.length) * colorArray.length))
```
brharrington, line 78: Is there a reason to use this instead of just the modulo operator? I think the operator makes it easier to read and more accurately conveys the intent. 

```
math.abs(i) % colorArray.length
```

atlas-chart/src/test/scala/com/netflix/atlas/chart/model/PaletteSuite.scala
```diff
@@ -35,20 +35,23 @@ class PaletteSuite extends FunSuite {
 
   test("colors:f00") {
     val p = Palette.create("colors:f00")
-    assertEquals(p.colors(0), Color.RED)
+    assertEquals(p.colorArray(0), Color.RED)
```
brharrington, line 38: I think the `colors` method could still be used here which would provide better coverage and the tests wouldn't need to change.

manolama: find and replace oops.

brharrington (approved):

PR #1511, opened 2023-01-13
Add the HEATMAP line style.

manolama:
Add the HeatmapDef class used to configure heatmaps globaly for a graph. Add a palette option to the LineDef class for use by heatmaps. Parse the HeatmapDef in Grapher.scala and pass it to the Axis and then PlotDef.
Throw an exception if a user attempts to plot a heatmap and stacked lines on the same axis.
Select a color or palette for use in heatmaps in Grapher.scala. If no style or heatmap settings are provided, the heatmap grabs a color based on the order of the expression and uses alpha transparencies for the color scale. Remaining expressions use the rest of the palette as normal. If the user specifies a color for the expression, that color is used with alphas and no colors are taken from the main palette. If a palette is specified for heatmaps, that palette is used for the color scale.

### DIFF
--- a/atlas-chart/src/main/scala/com/netflix/atlas/chart/model/DataDef.scala
+++ b/atlas-chart/src/main/scala/com/netflix/atlas/chart/model/DataDef.scala
@@ -52,6 +52,8 @@ sealed trait DataDef {
   *     Width of the stroke when rendering the line. Has no effect for styles other than LINE.
   * @param legendStats
   *     Summary stats for the data in the line.
+  * @param palette
+  *     An optional link to a palette override for this line..
   */
 case class LineDef(
   data: TimeSeries,
@@ -60,7 +62,8 @@ case class LineDef(
   color: Color = Color.RED,
   lineStyle: LineStyle = LineStyle.LINE,
   lineWidth: Float = 1.0f,
-  legendStats: SummaryStats = SummaryStats.empty
+  legendStats: SummaryStats = SummaryStats.empty,
+  palette: Option[Palette] = None
 ) extends DataDef {
 
   def label: String = data.label
--- a/atlas-chart/src/main/scala/com/netflix/atlas/chart/model/DataDef.scala
+++ b/atlas-chart/src/main/scala/com/netflix/atlas/chart/model/DataDef.scala
@@ -52,8 +52,6 @@ sealed trait DataDef {
   *     Width of the stroke when rendering the line. Has no effect for styles other than LINE.
   * @param legendStats
   *     Summary stats for the data in the line.
-  * @param palette
-  *     An optional link to a palette override for this line..
   */
 case class LineDef(
   data: TimeSeries,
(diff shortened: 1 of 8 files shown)


atlas-chart/src/main/scala/com/netflix/atlas/chart/model/DataDef.scala
```diff
@@ -60,7 +62,8 @@ case class LineDef(
   color: Color = Color.RED,
   lineStyle: LineStyle = LineStyle.LINE,
   lineWidth: Float = 1.0f,
-  legendStats: SummaryStats = SummaryStats.empty
+  legendStats: SummaryStats = SummaryStats.empty,
+  palette: Option[Palette] = None
```
brharrington, line 66: Why is this needed? I would have expected the palette for a heatmap to be resolved before this and would be set on the HeatmapDef.

atlas-eval/src/main/scala/com/netflix/atlas/eval/graph/Grapher.scala
```diff
@@ -295,6 +295,19 @@ case class Grapher(settings: DefaultSettings) {
...
+        if (
+          (dfltStyle == LineStyle.STACK || exprs.exists(
+            _.lineStyle.fold(dfltStyle)(s => LineStyle.valueOf(s.toUpperCase)) == LineStyle.STACK
+          )) && exprs.exists(
+            _.lineStyle
+              .fold(dfltStyle)(s => LineStyle.valueOf(s.toUpperCase)) == LineStyle.HEATMAP
+          )
+        ) {
+          throw new IllegalArgumentException(
+            "Mixing STACK and HEATMAP line styles on the same axis is not allowed."
```
brharrington, line 307: > Throw an exception if a user attempts to plot a heatmap and stacked lines on the same axis.

What is the rationale for this? While I would agree it is not generally going to be useful, I don't really see a reason to prohibit that combination.

brharrington (commented):
Added a few questions after a first pass, will take a closer look in a bit.

atlas-chart/src/main/scala/com/netflix/atlas/chart/model/DataDef.scala
manolama: I could do it that way, via heatmap def. I just have to modify the `Axis` in the `Grapher`. Should be easy.

atlas-eval/src/main/scala/com/netflix/atlas/eval/graph/Grapher.scala
manolama: For the percentile heatmap, I created a new ValueAxis type. If stacking is involved, the calculations pollute a lot of the heatmap logic (along with the axis) and I figured it wasn't worth it to support an op that could easily wind up completely obscuring the map. And if someone accidentally tried to plot percentiles or something else they wanted to line up with the map, it wouldn't so they'd be more confused. I think it's best to just ask the user to modify their query in that case.

brharrington:
Incorporated as part of #1547.

PR #1521, opened 2023-02-24
Atlas Akka - Set isSmile(), inputStream() and gzipMagicHeader public …

manolama:
…for re-use

in other projects.


manolama:
Actually not needed. Can just use the `customJson` method to extract JSON data.

PR #1522, opened 2023-02-24
Atlas Akka - Add an AkkaHttpClient for wrapping calls to enable easier

manolama:
unit testing.


atlas-akka/src/main/scala/com/netflix/atlas/akka/AkkaHttpClient.scala
```diff
@@ -0,0 +1,51 @@
...
+    * If an explicit [[ConnectionContext]] is given then it rather than the configured default [[ConnectionContext]] will be used
+    * for setting up the HTTPS connection pool, if the request is targeted towards an `https` endpoint.
+    *
+    * Note that the request must have an absolute URI, otherwise the future will be completed with an error.
+    */
+  def singleRequest(request: HttpRequest): Future[HttpResponse]
+
+}
+
+@Singleton
```
brharrington, line 45: The javax annotations should be removed. We are trying to phase that out to prepare for spring 6, e.g. #1482.

atlas-akka/src/main/scala/com/netflix/atlas/akka/AkkaHttpClient.scala
```diff
@@ -0,0 +1,51 @@
...
+class DefaultAkkaHttpClient @Inject() (implicit val system: ActorSystem) extends AkkaHttpClient {
+
+  override def singleRequest(request: HttpRequest): Future[HttpResponse] = {
+    Http().singleRequest(request)
```
brharrington, line 49: If this is going to be a common pattern, then it should probably setup the AccessLogger so it has consistent client metrics/logs.

brharrington (approved):

PR #1523, opened 2023-02-25
Atlas Core:

manolama:
Add some methods too the ArrayHelper object for dealing with byte arrays. Add setLong(3) and getLong(2) to to set and read longs from arbitrary positions in byte arrays.
Add memcmp(2) and memcmpMaybeNull(2) to evaluate the difference between the contents of two byte arrays.
Add byteArraysEquals(2) as a simple helper to determine if the length and content of two byte arrays are the same using memcmpMaybeNull(2). Borrowed from https://github.com/OpenTSDB/opentsdb/blob/9c1b775657254414e7c1b768c0f6804778ed6ff2/common/src/main/java/net/opentsdb/utils/Bytes.java.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/util/ArrayHelper.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/util/ArrayHelper.scala
@@ -17,6 +17,8 @@ package com.netflix.atlas.core.util
 
 import java.util.Comparator
 import scala.reflect.ClassTag
+import scala.util.control.Breaks.break
+import scala.util.control.Breaks.breakable
 
 object ArrayHelper {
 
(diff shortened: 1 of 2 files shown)


atlas-core/src/main/scala/com/netflix/atlas/core/util/ArrayHelper.scala
```diff
@@ -440,4 +442,129 @@ object ArrayHelper {
...
+    *
+    * @param a
+    *     First byte array to compare.
+    * @param b
+    *     Second byte array to compare
+    * @return
+    *     True if both arrays are null or have identical content and length. Otherwise
+    *     false.
+    */
+  def byteArraysEquals(a: Array[Byte], b: Array[Byte]): Boolean = memcmpMaybeNull(a, b) == 0
```
brharrington, line 458: Why not use [Arrays.equals(byte[], byte[])](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Arrays.html#equals(byte%5B%5D,byte%5B%5D))? They seem to have the same behavior and the built in can potentially benefit from things like intrinsics.

atlas-core/src/main/scala/com/netflix/atlas/core/util/ArrayHelper.scala
```diff
@@ -440,4 +442,129 @@ object ArrayHelper {
...
+    *
+    * @param a
+    *     First non-{@code null} byte array to compare.
+    * @param b
+    *     Second non-(@code null} byte array to compare
+    * @return
+    *     0 if the two arrays are identical, otherwise the difference between the
+    *     first two different bytes, otherwise the different between their lengths.
+    */
+  private def memcmp(a: Array[Byte], b: Array[Byte]): Int = {
```
brharrington, line 494: In jdk 9+ there is [Arrays.compareUnsigned(byte[], byte[])](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Arrays.html#compareUnsigned(byte%5B%5D,byte%5B%5D)). I think that should have the same ordering as used here.

atlas-core/src/main/scala/com/netflix/atlas/core/util/ArrayHelper.scala
```diff
@@ -440,4 +442,129 @@ object ArrayHelper {
...
+    // Private for now since we aren't using it for sorting.
+    val length = java.lang.Math.min(a.length, b.length)
+    if (a eq b) { // Do this after accessing a.length and b.length
+      return 0 // in order to NPE if either a or b is null.
+    }
+    var diff = 0
+    breakable {
```
brharrington, line 501: I would recommend avoiding `Breaks.breakable`. It is syntactic sugar to have something like the java `break` when using lambdas, but relies on exceptions to achieve the control flow which isn't particularly efficient.

A better option here would be to use a while loop instead of a for comprehension.

atlas-core/src/main/scala/com/netflix/atlas/core/util/ArrayHelper.scala
```diff
@@ -440,4 +442,129 @@ object ArrayHelper {
...
+    *
+    * @param b
+    *     A non-null byte array to write to with at least 8 bytes at the given
+    *     offset to store the long.
+    * @param n
+    *     The value to encode into the array.
+    * @param offset
+    *     An offset into the byte array.
+    */
+  def setLong(b: Array[Byte], n: Long, offset: Int): Unit = {
```
brharrington, line 525: Any reason not to just use [ByteBuffer](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/ByteBuffer.html)?

```java
ByteBuffer.wrap(b).putLong(offset, n)
```

atlas-core/src/main/scala/com/netflix/atlas/core/util/ArrayHelper.scala
```diff
@@ -440,4 +442,129 @@ object ArrayHelper {
...
+    *
+    * @param b
+    *     A non-null byte array to read from with at least 8 bytes at the given
+    *     offset to read the long.
+    * @param offset
+    *     An offset into the byte array.
+    * @return
+    *     The signed long read from the array.
+    */
+  def getLong(b: Array[Byte], offset: Int): Long = {
```
brharrington, line 554:
```java
ByteBuffer.wrap(b).getLong(offset)
```

atlas-core/src/main/scala/com/netflix/atlas/core/util/ArrayHelper.scala
manolama: Yeah I can use the new unsigned, thanks!

atlas-core/src/main/scala/com/netflix/atlas/core/util/ArrayHelper.scala
manolama: Oh ew, silly Scala. Thanks.

atlas-core/src/main/scala/com/netflix/atlas/core/util/ArrayHelper.scala
manolama: I'll use the ByteBuffer for now, have to verify the compiler will actually skip the object creation if our throughput grows.

PR #1533, opened 2023-03-27
eval: sum aggregation for gauges

brharrington:
Adds a new datapoint aggregator implementation to handle sum/count of gauge values that come from an aggregation service where there can be duplicates. The aggregation service would add a tag `atlas.aggr` with a key to use for grouping and deduping the relevant values. Values within a group will use max aggregation to be consistent with the behavior if there was a single instance of the aggregation service.

### DIFF
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
@@ -24,6 +24,7 @@ import com.netflix.atlas.core.model.Datapoint
 import com.netflix.atlas.core.model.Query
 import com.netflix.atlas.core.model.TimeSeries
 import com.netflix.atlas.core.util.Math
+import com.netflix.atlas.core.util.RefDoubleHashMap
 import com.netflix.spectator.api.Counter
 import com.netflix.spectator.api.Registry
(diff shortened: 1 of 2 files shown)


atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
```diff
@@ -172,20 +175,58 @@ object AggrDatapoint {
...
+      }
+      this
+    }
+
+    override def datapoints: List[AggrDatapoint] = List(datapoint)
+
+    def datapoint: AggrDatapoint = {
+      val tags = init.tags - aggrTagKey
+      var sum = 0.0
+      maxValues.foreach { (_, v) => sum = op(sum, v) }
```
manolama, line 208: I assume `NaN`s are filtered out before this point.

manolama (approved):

atlas-eval/src/main/scala/com/netflix/atlas/eval/model/AggrDatapoint.scala
brharrington: There shouldn't be `NaN` values here, though note the `op` in this case is `addNaN` so it can handle them in the expected way.

PR #1547, opened 2023-04-27
add support for heatmaps

brharrington:
Adds a heatmap line style that can be used to
render a set of lines on an axis as a heatmap
instead of a line. There can only be a single
heatmap on a particular axis. The `:heatmap`
operator can be used to select this style.

The cells for the heatmap are determined based
on the step size for the time axis and the tick
grid for the value axis. For normal lines each
line that overlaps a given cell will contribute
one to the count.

For percentile approximations that would used
with the `:percentiles` operator, they can now
be used with `:percentiles-heatmap` to map the
distribution to a heatmap. In this case the
count for a cell will be determined by the
distribution for the percentile approximation.

Additional axis specific parameters have been
added for `scale`, `u`, `l`, `palette`, and
`label` with a prefix of `heatmap_`. The scale
will control how counts are mapped to colors.

Since a heatmap can combine many lines, the
default legend test is just `Heatmap`. The
`heatmap_label` parameter on the URL can be
used to customize the label.


atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/Heatmap.scala
```diff
@@ -0,0 +1,260 @@
...
+  private val colorScale = Scales.factory(settings.colorScale)(
+    settings.lower.lower(hasArea = false, minCount),
+    settings.upper.upper(hasArea = false, maxCount),
+    palette.colorArray.size,
+    0
+  )
+
+  /** Set of ticks for the color scale used in legends. */
+  val colorTicks: ArraySeq[ValueTick] = {
+    val numTicks = palette.colorArray.size
```
manolama, line 101: TODO bug we can fix later, there are a couple of things with palettes.
1) If we have more colors in a custom palette than there are bucket ticks, we get an AIOOBE [here](https://github.com/Netflix/atlas/pull/1547/files#diff-1112a32851ca33315982b2b5de92a619efb6f2f3aa5ed2ca3a0093dba31683ecR231).
2) If there are, say 13 colors in the palette, then we just get the min and max but no values in between for the color legend. I have some code that can wrap the legend to show all the tick values if we want it.

atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/Heatmap.scala
```diff
@@ -0,0 +1,260 @@
...
+      val upper = math.min(mx, cellMax)
+      (upper - lower) / (mx - mn)
+    }
+  }
+
+  private def updateCounts(mn: Double, mx: Double, cnt: Double, counts: Array[Double]): Unit = {
+    var cellMin = minValue
+    var i = 0
+    while (i < yTicks.length) {
+      val cellMax = yTicks(i).v
```
manolama, line 164: I'd still like to add a percentiles scale that creates cells with a consistent height and ticks that align with the bucket boundaries. That's what most percentile heatmap implementations do so users are expecting that. The current scale with a weighting makes it a bit confusing, e.g. the [single percentile example](https://github.com/brharrington/atlas/pull/1/files#diff-2239050f6eed8ce2458d9f80316f64be58e53d0791b5d5d28764f60fad26e78c) has a smaller count for the top row and that makes it seem like more percentile buckets were returned than expected.
Users may also want a bit more fidelity around the lower latency bands. The linear scale [aggregates](https://github.com/Netflix/atlas/pull/1547/files#diff-f988b45359dd8a6079905ca462b32bbb701e6a5d184ad9a1004d92373a688dc0) a lot of data at the bottom while the buckets at the top are tall. [Log scale](https://github.com/brharrington/atlas/pull/1/files#diff-51e07e6fb9a390673fafc3f777fa68b094e43daf2d5e82ad1c4226adcecbb0bd) doesn't work well either.

atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/TimeSeriesHeatmap.scala
```diff
@@ -0,0 +1,59 @@
...
+      val px2 = xscale(t)
+
+      var i = 0
+      while (i < heatmap.numberOfValueBuckets) {
+        heatmap.color(t, i).foreach { c =>
+          val py1 = if (i == 0) y2 else yscale(heatmap.yTicks(i - 1).v)
+          val py2 = if (i == heatmap.yTicks.length) y1 else yscale(heatmap.yTicks(i).v)
+          // Use white base to allow gradients based on alpha to show up clearly if the
+          // overall background color is black
+          g.setColor(Color.WHITE)
```
manolama, line 48: ah nice catch.

manolama (approved):
So much cleaner than my code, thanks! And thanks for catching the dark background issues.

I still want to do something with a ptile scale and I have a couple of questions around the legend but we can start with this.

atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/Heatmap.scala
brharrington: That is just a copy paste bug in the json encoding, it should be using the label for the color tick there. Pushed a fix.

atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/Heatmap.scala
```diff
@@ -0,0 +1,260 @@
...
+
+  /** Palette for the colors associated with a count. */
+  val palette: Palette = settings.palette.getOrElse {
+    Palette.gradient(lines.head.color)
+  }
+
+  private val counts: Array[Array[Double]] = computeCounts()
+
+  /**
+    * Min and max count for the heatmap. The min will always be 0. The max will be the largest
```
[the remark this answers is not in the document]
brharrington: It is a possible improvement for the future, but risks some ambiguity. In general for the lower/upper bounds on the heatmap I was a bit torn. One option that would be fairly consistent with the way it works on an axis would be to simply ignore cells that exceed the bounds. However, since it is visible that could be confused with no data for that interval which isn't really an issue with the normal graph bounds. So currently it uses a clamping behavior and essentially maps values that exceed a bound to that bound.

atlas-chart/src/main/scala/com/netflix/atlas/chart/graphics/Heatmap.scala
brharrington: We can think about it for the future, but it is quite messy. The spacing is not uniform and was chosen to be mathematically convenient to compute the bucket and give a reasonable approximation, but that doesn't translate to being easy for a user to reason about. None of the ticks would be on convenient boundaries and the non uniform spacing means you couldn't really omit values to have a user reason about it so the axis gets messy. I think we can see how usage goes and go from there.

PR #1548, opened 2023-04-29
add log-linear scale type

brharrington:
It behaves like a log scale for powers of 10 with
linear behavior in between. This helps spread out
the smaller values. This can be useful for things
like a heatmap view of percentile distributions.


manolama (approved):
Yeah that one looks better. Thanks.

PR #1610, opened 2024-02-15
atlas JMH: Quantify skipping decoding of the time series ID

manolama:
when decoding batches of datapoints.


atlas-jmh/src/main/scala/com/netflix/atlas/webapi/PublishPayloadsBench.scala
```diff
@@ -126,6 +133,14 @@ class PublishPayloadsBench {
     }
   }
 
+  @Benchmark
+  def decodeBatchDatapoints(bh: Blackhole): Unit = {
```
brharrington, line 137: The comment at the top should be updated so there is a reference point for the behavior we are seeing.

brharrington (approved):

PR #1658, opened 2024-05-21
Update README.md

jhona32423:
Fixing a small typo at the README


brharrington:
That isn't a typo, it is upper case in reference to `(the “License”)` earlier.

PR #1675, opened 2024-07-30
eval: add a rewrite module to handle queries with namespaces.

manolama:
DataSource URIs are sent via HTTP (with retries) and if they incorporate namespace predicates, the service will rewrite them to route to the proper host, removing the namespace predicate.

### DIFF
--- a/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
+++ b/atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
@@ -60,6 +60,7 @@ import com.netflix.atlas.eval.stream.Evaluator.MessageEnvelope
 import com.netflix.atlas.json.Json
 import com.netflix.atlas.json.JsonSupport
 import com.netflix.atlas.pekko.ClusterOps
+import com.netflix.atlas.pekko.DiagnosticMessage
 import com.netflix.atlas.pekko.StreamOps
 import com.netflix.atlas.pekko.ThreadPools
 import com.netflix.spectator.api.Registry
@@ -86,8 +87,11 @@ private[stream] abstract class EvaluatorImpl(
 
   private val logger = LoggerFactory.getLogger(getClass)
 
+  // Calls out to a rewrite service in case URIs need mutating to pick the proper backend.
+  private[stream] var dataSourceRewriter = new DataSourceRewriter(config, system)
+
   // Cached context instance used for things like expression validation.
-  private val validationStreamContext = newStreamContext()
+  private val validationStreamContext = newStreamContext(new ThrowingDSLogger)
 
   // Timeout for DataSources unique operator: emit repeating DataSources after timeout exceeds
   private val uniqueTimeout: Long = config.getDuration("atlas.eval.stream.unique-timeout").toMillis
@@ -129,7 +133,13 @@ private[stream] abstract class EvaluatorImpl(
   }
 
   protected def validateImpl(ds: DataSource): Unit = {
-    validationStreamContext.validateDataSource(ds).get
+    val future = Source
+      .single(DataSources.of(ds))
+      .via(dataSourceRewriter.rewrite(validationStreamContext))
+      .map(_.sources().asScala.map(validationStreamContext.validateDataSource).map(_.get))
+      .toMat(Sink.head)(Keep.right)
+      .run()
+    Await.result(future, 60.seconds)
   }
 
   protected def writeInputToFileImpl(uri: String, file: Path, duration: Duration): Unit = {
@@ -212,6 +222,7 @@ private[stream] abstract class EvaluatorImpl(
   def createStreamsFlow: Flow[DataSources, MessageEnvelope, NotUsed] = {
     val (logSrc, context) = createStreamContextSource
     Flow[DataSources]
+      .via(dataSourceRewriter.rewrite(context))
       .map(dss => groupByHost(dss))
       // Emit empty DataSource if no more DataSource for a host, so that the sub-stream get the info
       .via(new FillRemovedKeysWith[String, DataSources](_ => DataSources.empty()))
@@ -580,5 +591,17 @@ private[stream] abstract class EvaluatorImpl(
 
   private def isPrintable(c: Int): Boolean = {
     c >= 32 && c < 127
+  }
+
+  private class ThrowingDSLogger extends DataSourceLogger {
+
+    override def apply(ds: DataSource, msg: JsonSupport): Unit = {
+      msg match {
+        case dsg: DiagnosticMessage =>
+          throw new IllegalArgumentException(dsg.message)
+      }
+    }
+
+    override def close(): Unit = {}
   }
 }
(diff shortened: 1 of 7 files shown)


atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/NamespaceRewriter.scala
```diff
@@ -0,0 +1,251 @@
...
+                    promise.failure(
+                      new RuntimeException(s"Failed to rewrite URIs, status: $body")
+                    )
+                  case Failure(e) =>
+                    logger.error("Failed to read response body", e)
+                    promise.failure(e)
+                }
+            case _ => failWithBody(promise, response)
+          }
+        case Failure(ex) =>
```
brharrington, line 209: This is typically handled with settings on the pool:

https://pekko.apache.org/docs/pekko-http/current/client-side/host-level.html#retrying-a-request

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/NamespaceRewriter.scala
```diff
@@ -0,0 +1,251 @@
...
+    client: PekkoHttpClient
+  ): Future[DataSources] = {
+    if (!config.hasPath("rewriteUrl") || ds.sources().isEmpty) {
+      return Future.successful(ds)
+    }
+
+    implicit val ec = system.dispatcher
+    val promise = Promise[DataSources]()
+    val dataSources = ds.sources().asScala.toList
+    val uri = config.getString("rewriteUrl")
```
brharrington, line 76: This should be prefixed. For HOCON hyphen separated is the recommendation:

https://github.com/lightbend/config/blob/main/HOCON.md#hyphen-separated-vs-camelcase

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/NamespaceRewriter.scala
```diff
@@ -0,0 +1,251 @@
...
+import scala.concurrent.duration.FiniteDuration
+import scala.jdk.CollectionConverters.CollectionHasAsScala
+import scala.jdk.CollectionConverters.SetHasAsJava
+import scala.util.Failure
+import scala.util.Random
+import scala.util.Success
+import scala.util.Try
+import scala.util.Using
+
+class NamespaceRewriter(config: Config, implicit val system: ActorSystem) extends StrictLogging {
```
brharrington, line 50: In theory we could use it for other things, not sure it should be specific to namespaces. It is just an endpoint we use to perform a rewrite on the URIs.

brharrington (changes requested):

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/EvaluatorImpl.scala
```diff
@@ -194,6 +198,7 @@ private[stream] abstract class EvaluatorImpl(
     */
   def createStreamsFlow: Flow[DataSources, MessageEnvelope, NotUsed] = {
     Flow[DataSources]
+      .mapAsync(1)(namespaceRewriter.rewriteNamespaces)
```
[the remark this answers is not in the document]
manolama: I'll give that a try.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/NamespaceRewriter.scala
```diff
@@ -0,0 +1,251 @@
...
+  ): Try[DataSources] = {
+    var i = 0
+    val rewrites = List.newBuilder[DataSource]
+    Using(Json.newJsonParser(new ByteStringInputStream(body))) { parser =>
+      foreachItem(parser) {
+        var newUri: String = null
+        var ok = false
+        var msg: String = null
+        var original: String = null
+        foreachField(parser) {
```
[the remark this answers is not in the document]
manolama: I was worried about huge lists of URIs but we'll likely only have one or two at a time so a class would be fine.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/NamespaceRewriter.scala
```diff
@@ -0,0 +1,251 @@
...
+          case "status" =>
+            val status = parser.nextTextValue()
+            if (status.toUpperCase() == "OK") {
```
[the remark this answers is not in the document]
manolama: It is, just being defensive for the future.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/NamespaceRewriter.scala
```diff
@@ -0,0 +1,251 @@
...
+          val ds = dataSources(i)
+          rewrites += new DataSource(
+            ds.id,
+            ds.step(),
+            newUri
+          )
+        } else {
+          logger.error("Failed to rewrite URI, message: {}", msg)
+          return Failure(
+            new IllegalArgumentException(s"Failed to rewrite URI $original, message: $msg")
```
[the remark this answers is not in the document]
manolama: Drats, right.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/NamespaceRewriter.scala
manolama: Ah that's how to do it, thanks.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/NamespaceRewriter.scala
manolama: It is prefixed with `atlas.eval.graph` but it's stripped out by `EvaluatorImpl`

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/NamespaceRewriter.scala
manolama: ack

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/NamespaceRewriter.scala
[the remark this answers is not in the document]
brharrington: It might be thousands in some cases, but still I don't think this would be the hot spot and clarity and maintainability are probably better tradeoffs here.

atlas-eval/src/main/scala/com/netflix/atlas/eval/stream/NamespaceRewriter.scala
brharrington: #1677 updates the helper to have a flow option and instrument it better.

I'll have to think about the degradation a bit. Will need to ensure that the rewrite call failing overall will have no impact other than preventing data source changes and that those would get picked up automatically when it recovers.

manolama: Caching the rewrites for a reconnect? That's the only time we'd need to worry about it right? Otherwise if a new eval was submitted the caller would need to retry.

brharrington (approved):

PR #1690, opened 2024-08-22
Override SourceQueue.offer to accept Tags

JulianGriggs:
Adds the ability for a user to specify additional tags to be attached to queue metrics

### DIFF
--- a/atlas-pekko/src/main/scala/com/netflix/atlas/pekko/StreamOps.scala
+++ b/atlas-pekko/src/main/scala/com/netflix/atlas/pekko/StreamOps.scala
@@ -37,6 +37,7 @@ import org.apache.pekko.stream.stage.OutHandler
 import org.apache.pekko.stream.stage.TimerGraphStageLogic
 import com.netflix.spectator.api.Clock
 import com.netflix.spectator.api.Registry
+import com.netflix.spectator.api.Tag
 import com.typesafe.scalalogging.StrictLogging
 import org.apache.pekko.stream.stage.GraphStageWithMaterializedValue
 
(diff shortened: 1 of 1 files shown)


atlas-pekko/src/main/scala/com/netflix/atlas/pekko/StreamOps.scala
```diff
@@ -99,23 +100,41 @@ object StreamOps extends StrictLogging {
...
+     * Add the value into the queue if there is room. Returns true if the value was successfully
+     * enqueued.
+     */
     def offer(value: T): Boolean = {
       queue.offer(value) match {
-        case QueueOfferResult.Enqueued    => enqueued.increment(); true
-        case QueueOfferResult.Dropped     => dropped.increment(); false
-        case QueueOfferResult.QueueClosed => closed.increment(); false
-        case QueueOfferResult.Failure(_)  => failed.increment(); false
+        case QueueOfferResult.Enqueued    => registry.counter(enqueuedId).increment(); true
```
brharrington, line 116: Putting data into the queue is often a hot path, adding the counter lookup for every call is probably not desirable.

brharrington (commented):
What is the use-case for this change? It seems like if more detail is needed it could be done on a custom metric and limit the overhead to that use-case.

JulianGriggs:
The items that I'm enqueuing are coming from a number of different sources, each of which reports at a different rate (in this case, each source is an Alert object). It would be helpful to add the name of the source as a dimension to the queue metrics so that when there is a big jump in say `enqueued`, we can see where that jump came from. I can add a separate custom metric to get this same information, but as I was already leveraging these metrics (in a dashboard) I figured it would be reasonable to add this functionality. But I agree that this could lead to perf issues in certain cases so feel free to reject this change.

brharrington:
Thanks for the use-case description. Given there is an easy work around will close for now to reduce performance risk here.

PR #1692, opened 2024-09-05
core: support sub-second step sizes

brharrington:
Update step rounding logic to allow for sub-second step sizes.


manolama (approved):
Thanks.

PR #1693, opened 2024-09-20
chart: add default theme to JsonCodec parsing.

manolama:
The old scala 2.12 code did not serialize the theme in the v2.json graph metadata. For Core, they're using Spark to emit data in V2 then parsing it with the latest Atlas code. This will fix the issue for them.

### DIFF
--- a/atlas-chart/src/test/scala/com/netflix/atlas/chart/JsonCodecSuite.scala
+++ b/atlas-chart/src/test/scala/com/netflix/atlas/chart/JsonCodecSuite.scala
@@ -16,12 +16,15 @@
 package com.netflix.atlas.chart
 
 import com.netflix.atlas.chart.model.GraphDef
+import com.netflix.atlas.chart.model.Layout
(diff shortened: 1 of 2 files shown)


atlas-chart/src/test/scala/com/netflix/atlas/chart/JsonCodecSuite.scala
```diff
@@ -35,9 +38,51 @@ class JsonCodecSuite extends FunSuite {
...
+    assertEquals(gdef.legendType.toString, "LABELS_WITH_STATS")
+    assertEquals(gdef.onlyGraph, false)
+    assertEquals(gdef.warnings, List.empty)
+    assertEquals(gdef.themeName, "light")
+  }
+
   test("rendering hint: none") {
     val gdef = graphDef(Set.empty)
     val str = JsonCodec.encode(gdef)
+    System.out.println(str)
```
brharrington, line 85: The println should get removed.

PR #1897, opened 2026-04-13
Implement loadTestConfig for external test configurations

sand339:
Added loadTestConfig method to load external test configurations, with a note on potential path traversal vulnerabilities.


sand339:
Submitted by error

PR #1898, opened 2026-04-21
tag atlas.eval.datapoints counters with streamName

manolama:
Adds an optional streamName to Evaluator so apps that run multiple LWC evaluators (e.g. against different LWC APIs) can distinguish which stream produced buffered, dropped-old, dropped-future, or dropped-datapoints-limit-exceeded counts. Defaults to "default" so existing callers are unaffected.


brharrington:
Closing in favor of #1899

PR #1936, opened 2026-06-23
core: prevent a zero step size in Step.compute (and map ArithmeticException to 400)

batam111:
## Problem
A graph request whose `step` is much larger than the time window — e.g.
`/api/v1/graph?q=1&step=1000w&s=2012-01-01T00:00&e=2012-01-01T00:01` — returns **HTTP 500**
(`ArithmeticException: / by zero`).

## Root cause
When the requested window is smaller than the (rounded) primary step, `Step.compute` computes
`datapoints = (end - start) / primary = 0`. `datapointsPerPixel(0, width)` then returns 0, so
`minStep` is 0 and the resolved step size becomes 0. That zero step is later used as a divisor
(rounding timestamps to step boundaries), raising `ArithmeticException`, which the request handler
does not map and so surfaces as a 500.

## Fix
Two parts, addressing cause and symptom:
1. `Step.compute`: require at least one datapoint per pixel (`math.max(1L, datapointsPerPixel(...))`)
   so the resolved step can never round down to zero. This only changes the degenerate
   window-smaller-than-step case; for all other inputs `datapointsPerPixel` is already >= 1.
2. `RequestHandler.errorResponse`: also map `ArithmeticException` to `BadRequest` as defense in
   depth, so any future divide-by-zero is a client error rather than a 500.

With the clamp, such requests now return **HTTP 200** with a single coarse bucket (the window is
smaller than one step) instead of failing.

## Tests
`StepSuite` adds a regression for the window-smaller-than-step case and a check that `compute`
returns a positive step across a range of primaries / widths / windows.

## Note
The `RequestHandler` change touches the same line as a separate change adding
`UnsupportedOperationException`. If both land, the second is a one-line rebase.

### DIFF
--- a/atlas-core/src/main/scala/com/netflix/atlas/core/util/Step.scala
+++ b/atlas-core/src/main/scala/com/netflix/atlas/core/util/Step.scala
@@ -41,7 +41,10 @@ class Step private (allowedStepSizesForBlock: List[Long]) {
     */
   def compute(primary: Long, width: Int, start: Long, end: Long): Long = {
     val datapoints = (end - start) / primary
-    val minStep = datapointsPerPixel(datapoints, width) * primary
+    // Ensure at least one datapoint per pixel so the resulting step can never round down to
+    // zero. When the window is smaller than the primary step `datapoints` is 0, which would
+    // otherwise make `minStep` 0 and yield a zero step size (later used as a divisor).
+    val minStep = math.max(1L, datapointsPerPixel(datapoints, width)) * primary
     round(primary, minStep)
   }
 }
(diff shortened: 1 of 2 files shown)


atlas-pekko/src/main/scala/com/netflix/atlas/pekko/RequestHandler.scala
```diff
@@ -203,7 +203,7 @@ object RequestHandler extends StrictLogging {
     // Determine most appropriate status code to use based on the exception type
     t match {
       case e @ (_: IllegalArgumentException | _: IllegalStateException | _: JacksonException |
-          _: DateTimeException) =>
+          _: DateTimeException | _: ArithmeticException) =>
```
brharrington: I think in most cases, like the one shown here, this would indicate a server side bug so 500 is accurate.

